PR #2, opened 2017-12-20
feature/GE-636: Generated project, shifted settings, added pinned req…

AltusBarry:
…uirements, added apps, 2fa urls giving issues atm


project/settings_base.py
```diff
@@ -0,0 +1,120 @@
...
+
+# Build paths inside the project like this: os.path.join(BASE_DIR, ...)
+BASE_DIR = os.path.dirname(os.path.dirname(os.path.abspath(__file__)))
+
+
+# Quick-start development settings - unsuitable for production
+# See https://docs.djangoproject.com/en/1.11/howto/deployment/checklist/
+
+# SECURITY WARNING: keep the secret key used in production secret!
+SECRET_KEY = '_n(_w(3!i4-p((jz8(o0fb*_r5fb5t!qh1g@m9%4vryx5lale='
```
cobusc: I presume we will read this from the environment when running on containers?

cobusc:
:+1:

project/settings_base.py
AltusBarry: Ah yes, that one should be updated. Still a bit used to server local settings.

AltusBarry: Base setting left as is, settings.py has env override.

PR #5, opened 2018-01-10
Feature/ge 702 login page

ruands:
Credential and OTP wizard/forms.


core_authentication_service/forms.py
```diff
@@ -0,0 +1,42 @@
+from django import forms
+from django.contrib.auth import authenticate
+from django.contrib.auth.models import User
+from django.core.exceptions import ValidationError
+from django.utils.translation import ugettext as _
+
+
+class LoginForm(forms.Form):
```
AltusBarry: Know its a pain, sorry. Don't you wanna check if `from django.contrib.auth.forms import AuthenticationForm` is easier to subclass than needing to clean the passwords and stuff ourselves.

AltusBarry: Probably has some other niceties too.

core_authentication_service/forms.py
```diff
@@ -0,0 +1,42 @@
...
+            raise ValidationError(_("Please enter your password."))
+        return password
+
+    def clean(self):
+        username = self.cleaned_data.get("username")
+        password = self.cleaned_data.get("password")
+
+        if User.objects.filter(username=username).exists():
+            self.user_cache = authenticate(
+                self.request, username=username, password=password)
```
AltusBarry: Preferable to do anything in regards to the authentication and request usage in the view.

core_authentication_service/templates/base.html
```diff
@@ -0,0 +1,8 @@
+<html lang="en">
```
AltusBarry: html
    body
        main
            block

AltusBarry: Damnit github!

core_authentication_service/templates/core-authentication-service/login.html
```diff
@@ -0,0 +1,18 @@
...
+{% block content %}
+    <form method="post">
+        {% csrf_token %}
+        {{ wizard.management_form }}
+        {{ form.as_p }}
+        <button type="submit">Login</button>
+
+        {% if backup_tokens %}
+            <p>{% trans "As a last resort, you can use a backup token:" %}</p>
+            <p>
```
AltusBarry: Why button in a paragraph tag?

core_authentication_service/views.py
```diff
@@ -0,0 +1,114 @@
...
+from two_factor.forms import AuthenticationTokenForm, BackupTokenForm
+
+from core_authentication_service import forms
+
+
+class LoginView(IdempotentSessionWizardView):
+    """This view handles the login process. The first step asks for the user's
+    username and password. If the username and password is correct, it will
+    check if the user has 2FA enabled. If enabled, the second step is displayed
+    that requires the user to enter a OTP generated by a device (phone in this
```
AltusBarry: `an OTP`

AltusBarry: Specifically google authenticator, might not be solely phone bound.

core_authentication_service/views.py
```diff
@@ -0,0 +1,114 @@
...
+
+    Adapted from two_factor.views.core"""
+    template_name = "core-authentication-service/login.html"
+    form_list = (
+        ("credentials", forms.LoginForm),
+        ("token", AuthenticationTokenForm),
+        ("backup", BackupTokenForm),
+    )
+
+    idempotent_dict = {
```
AltusBarry: What does this dict do?

core_authentication_service/views.py
```diff
@@ -0,0 +1,114 @@
...
+    }
+
+    def __init__(self, **kwargs):
+        super(LoginView, self).__init__(**kwargs)
+        self.user_cache = None
+        self.device_cache = None
+
+    def done(self, form_list, **kwargs):
+        """Log user in and direct them back to the site they came from."""
+        login(self.request, self.get_user())
```
AltusBarry: Shouldn't oidc be in control of all this, in some form?

AltusBarry: Can this view be hit without oidc authorization? If so, we can not redirect outside this domain or on querystring redirect_uri unless oidc is happy with it.

core_authentication_service/views.py
```diff
@@ -0,0 +1,114 @@
...
+        device = getattr(self.get_user(), 'otp_device', None)
+        if device:
+            signals.user_verified.send(sender=__name__, request=self.request,
+                                       user=self.get_user(), device=device)
+        return redirect(redirect_to)
+
+    def get_form_kwargs(self, step=None):
+        if step == "credentials":
+            return {
+                "request": self.request
```
AltusBarry: Check how wizards handle form validation redirects. The wizard will still need to redirect to the current page with the full form and errors somehow.

Step clean probably.

core_authentication_service/views.py
```diff
@@ -0,0 +1,114 @@
...
+               "token" not in self.storage.validated_step_data
+
+    condition_dict = {
+        "token": has_token_step,
+        "backup": has_backup_step,
+    }
+
+    def __init__(self, **kwargs):
+        super(LoginView, self).__init__(**kwargs)
+        self.user_cache = None
```
AltusBarry: Don't make use of self like this on wizards.

core_authentication_service/views.py
```diff
@@ -0,0 +1,114 @@
...
+                "request": self.request
+            }
+        if step in ("token", "backup"):
+            return {
+                "user": self.get_user(),
+                "initial_device": self.get_device(step)
+            }
+        return {}
+
+    def get_device(self, step=None):
```
AltusBarry: Make get_device and get_user cached_properties.

core_authentication_service/views.py
AltusBarry: I can't really figure out the need for the "cached" values.

core_authentication_service/views.py
```diff
@@ -0,0 +1,114 @@
...
+            self.device_cache = default_device(self.get_user())
+        if step == "backup":
+            try:
+                self.device_cache = self.get_user().staticdevice_set.get(
+                    name="backup")
+            except StaticDevice.DoesNotExist:
+                pass
+        return self.device_cache
+
+    def get_user(self):
```
AltusBarry: Cached property won't help here as the value can change. However, I don't see the need for the user_cache variable, the processing power being saved is minimal with that if and its not helping to reduce db hits.

core_authentication_service/views.py
```diff
@@ -0,0 +1,114 @@
...
+        """Return the authenticated user or None if user isn't authenticated."""
+        if not self.user_cache:
+            form_obj = self.get_form(
+                step="credentials",
+                data=self.storage.get_step_data("credentials")
+            )
+            self.user_cache = form_obj.is_valid() and form_obj.user_cache
```
AltusBarry: Authenticate the user some way on the view where it can either be a cached property or stored on session storage.

core_authentication_service/views.py
ruands: From https://github.com/juanifioren/django-oidc-provider/blob/v0.5.x/example_project/ it seems that oidc refers to our logic for handling login and redirect.

core_authentication_service/views.py
```diff
@@ -0,0 +1,120 @@
...
+from django_otp.plugins.otp_static.models import StaticDevice
+
+from two_factor import signals
+from two_factor.utils import default_device
+from two_factor.views.utils import IdempotentSessionWizardView
+from two_factor.forms import AuthenticationTokenForm, BackupTokenForm
+
+
+class LoginView(IdempotentSessionWizardView):
+    """This view handles the login process. The first step asks for the user's
```
cobusc: This docstring is beautiful.

core_authentication_service/views.py
cobusc: https://github.com/Bouke/django-two-factor-auth/blob/master/two_factor/views/utils.py#L46-L54

I agree that this needs a bit of a comment. Although the name is `idempotent_dict` its use seems to suggest that one uses it to identify `non-idempotent` wizard steps, i.e. steps that may only be executed once.

core_authentication_service/views.py
cobusc: @ruands is it required to call `login()` from `django.contrib.auth`?

core_authentication_service/views.py
```diff
@@ -0,0 +1,120 @@
...
+        self.user_cache = None
+        self.device_cache = None
+
+    def done(self, form_list, **kwargs):
+        """Log user in and direct them back to the site they came from."""
+        login(self.request, self.get_user())
+
+        redirect_to = self.request.POST.get(
+            "next",
+            self.request.GET.get("next", '')
```
cobusc: A mix of double- and single quotes are used.

core_authentication_service/views.py
ruands: As discussed, `login()` is now only called when user is a superuser.

core_authentication_service/views.py
ruands: Tried using session storage. Issue is that it can't store an object. Trying to serialize the object so it can be stored as json failed due to a `datetime` field on the object. And just storing as a dict takes away the ability to call attributes such as `is_anonymous`.

One solution was to just store the object's id. This works for the user but for devices it does not work since there are 2 types of device objects. If we only store the id of the device, we have no way of knowing what type of device we need to return.

cobusc: @ruands @AltusBarry I see that `user_cache` and `device_cache` is fairly integral in the 2FA LoginView here: https://github.com/Bouke/django-two-factor-auth/blob/master/two_factor/views/core.py#L88-L89
Since there are multiple places in the construction of the wizard where the user is required, these `_cache` values prevent repeated DB calls. So this is not a cache that lives between wizard pages, just a mechanism to avoid hitting the DB again if we already have a value in the construction of the wizard. In that sense the use of these variables seems warranted.

core_authentication_service/views.py
```diff
@@ -0,0 +1,114 @@
...
+from two_factor import signals
+
+from two_factor.utils import default_device
+from two_factor.views.utils import IdempotentSessionWizardView
+from two_factor.forms import AuthenticationTokenForm, BackupTokenForm
+
+from core_authentication_service import forms
+
+
+class LoginView(IdempotentSessionWizardView):
```
cobusc: Would it not make more sense to rather just derive from this class: https://github.com/Bouke/django-two-factor-auth/blob/master/two_factor/views/core.py#L51
and then override the methods we need to?

core_authentication_service/tests/test_wizard.py
```diff
@@ -0,0 +1,159 @@
...
+        )
+        return response
+
+    def post_backup_step(self):
+        print("Posting backup step")
+        response = self.client.post(
+            self.wizard_url,
+            {
+                "login_view-current_step": "backup",
+                "backup-otp_token": "abcdef123",
```
AltusBarry: The token and passwords in the post methods, rather make them args passed in as well. So that if needed we can easily update the values without needing to find multiple places to do so.

core_authentication_service/tests/test_wizard.py
```diff
@@ -0,0 +1,159 @@
...
+    """Test case for users without 2FA enabled."""
+    def test_it(self):
+        # Get credentials step
+        response = self.get_credential_step()
+        self.assertEqual(response.status_code, 200)
+        self.assertContains(response, "Username")
+
+        # Post credentials step. We check for a username field in the response
+        # since only staff memebers will actually have access to admin, other
+        # users will see another login screen.
```
AltusBarry: So we only have tests for staff users? The two test users didn't get is_staff set.

core_authentication_service/tests/test_wizard.py
```diff
@@ -0,0 +1,159 @@
...
+            {"__all__":
+                ["Invalid token. Please make sure you have entered it "
+                 "correctly."]
+             }
+        )
+
+        # Post backup step
+        response = self.post_backup_step()
+        self.assertEqual(response.status_code, 200)
+        self.assertRedirects(response, "/login/?next=/admin/")
```
AltusBarry: All the tests end up with this url. Should some of them not redirect back to oidc clients?

core_authentication_service/views.py
```diff
@@ -0,0 +1,33 @@
...
+
+    template_name = "core-authentication-service/login.html"
+
+    def done(self, form_list, **kwargs):
+        if self.get_user().is_superuser:
+            login(self.request, self.get_user())
+
+        redirect_to = self.request.POST.get(
+            "next",
+            self.request.GET.get("next", '')
```
AltusBarry: Double and single quotes mixed.

core_authentication_service/views.py
```diff
@@ -0,0 +1,33 @@
...
+        )
+        if not is_safe_url(url=redirect_to, host=self.request.get_host()):
+            redirect_to = resolve_url(settings.LOGIN_REDIRECT_URL)
+
+        device = getattr(self.get_user(), 'otp_device', None)
```
AltusBarry: Use double quotes.

core_authentication_service/tests/test_wizard.py
ruands: No I test regular users, since they will go through the same flow, regardless of whether the login occurs. I will add a test specifically for staff/superusers.

core_authentication_service/tests/test_wizard.py
ruands: That is the default url for login redirect. OIDC takes care of the redirect back to the client. I don't test that.

AltusBarry: Cool, good to know.

AltusBarry:
+1, just check the conflicts I introduced into dev.

PR #6, opened 2018-01-16
feature/GE-703

AltusBarry:
Registration for auth service.


README.rst
```diff
@@ -2,3 +2,5 @@
 
 Documentation for this project can be found here:
 https://girleffect.github.io/core-authentication-service/
+
+Migration 0001 for otp_totp will break if user AUTH_USER_MODEL is changed after running migrations.
```
cobusc: Thanks for adding this

core_authentication_service/admin.py
```diff
@@ -0,0 +1,33 @@
...
+
+from core_authentication_service import models
+
+
+@admin.register(models.CoreUser)
+class CoreUserAdmin(UserAdmin):
+    fieldsets = UserAdmin.fieldsets + (
+        ("Extra data",
+            {"fields": (
+                "nickname", "msisd", "birth_date", "country_code", "avatar"
```
cobusc: Typo: "msisd"

core_authentication_service/forms.py
```diff
@@ -0,0 +1,193 @@
...
+            "username", "first_name", "last_name", "email",
+            "nickname", "msisdn", "gender", "birth_date", "country", "avatar"
+        ]
+
+    def __init__(self, security=None, required=[],*args, **kwargs):
+        # Security value is required later in form processes as well.
+        self.security = security
+
+        # Super needed before we can actually update the form.
+        super(RegistrationForm, self).__init__(*args, **kwargs)
```
cobusc: I would call super before starting to modify the instance, i.e. at the top of the function

core_authentication_service/forms.py
```diff
@@ -0,0 +1,193 @@
...
+
+        # Set update form update variables, for manipulation as init
+        # progresses.
+        required_fields = []
+        fields_data = {}
+        if required:
```
cobusc: Not necessary to check this. If the required list is empty, then the following line will do nothing.

core_authentication_service/forms.py
```diff
@@ -0,0 +1,193 @@
...
+            for field in required:
+                required_fields += REQUIREMENT_DEFINITION.get(field, [field])
```
cobusc: For interest's sake, one can also do something like this here:
```python
required_fields = list(itertools.chain.from_iterable(
    REQUIREMENT_DEFINITION.get(field, [field]) for field in required
))
```

core_authentication_service/forms.py
```diff
@@ -0,0 +1,193 @@
...
+
+        # Security level needs some additions before the form is rendered.
+        if self.security == "high":
+            required_fields += ["email"]
+        else:
```
cobusc: Why is there an `else` here? It is not clear why the password1 help_text attribute is set to empty if the security level is not `high`.

core_authentication_service/forms.py
```diff
@@ -0,0 +1,193 @@
...
+    def __init__(self, security=None, required=[],*args, **kwargs):
+        # Security value is required later in form processes as well.
+        self.security = security
+
+        # Super needed before we can actually update the form.
+        super(RegistrationForm, self).__init__(*args, **kwargs)
+
+        # Set update form update variables, for manipulation as init
+        # progresses.
+        required_fields = []
```
cobusc: This can be change to a `set`

core_authentication_service/forms.py
```diff
@@ -0,0 +1,193 @@
...
+                    }
+                }
+            }
+
+        incorrect_fields = set(required_fields) - set(self.fields.keys())
+        for field in incorrect_fields:
+            LOGGER.warning(
+                "Received field to alter that is not on form: %s" % field
+            )
+            while required_fields.count(field) > 0:
```
cobusc: If `required_fields` was a set, this would be cleaner (and more efficient).

core_authentication_service/forms.py
```diff
@@ -0,0 +1,193 @@
...
+                        "help_text": ""
+                    }
+                }
+            }
+
+        incorrect_fields = set(required_fields) - set(self.fields.keys())
+        for field in incorrect_fields:
+            LOGGER.warning(
+                "Received field to alter that is not on form: %s" % field
+            )
```
cobusc: Should we not rather raise an exception if there are incorrect fields? These are programmer errors which we cannot allow.

core_authentication_service/forms.py
```diff
@@ -0,0 +1,193 @@
...
+        if password1 and password2 and password1 != password2:
+            raise forms.ValidationError(
+                self.error_messages['password_mismatch'],
+                code='password_mismatch',
+            )
+        self.instance.username = self.cleaned_data.get('username')
+
+        # NOTE: Min length might need to be defined somewhere easier to change.
+        # Setting doesn't feel 100% right though.
+        if not len(password1) > 3 and not len(password2) > 3:
```
cobusc: Noted your comment. At least create a constant in this module so that there are no direct references to `3`.

cobusc: If we have already checked that the passwords are the same (implying equal length), it is not necessary to check the length of both.

core_authentication_service/forms.py
```diff
@@ -0,0 +1,193 @@
...
+            return super(RegistrationForm, self).clean_password2()
+
+        password1 = self.cleaned_data.get("password1")
+        password2 = self.cleaned_data.get("password2")
+        if password1 and password2 and password1 != password2:
+            raise forms.ValidationError(
+                self.error_messages['password_mismatch'],
+                code='password_mismatch',
+            )
+        self.instance.username = self.cleaned_data.get('username')
```
cobusc: Why does the `username` on the instance get set in the `clean_password2` function?

core_authentication_service/forms.py
```diff
@@ -0,0 +1,193 @@
...
+        if not email and not msisdn:
+            raise ValidationError("Enter either email or msisdn")
+
+        return super(RegistrationForm, self).clean()
+
+
+class SecurityQuestionFormSetClass(BaseFormSet):
+    def __init__(self, language, *args, **kwargs):
+        self.language = language
+        super(SecurityQuestionFormSetClass, self).__init__(*args, **kwargs)
```
cobusc: My preference would be to call super first before modifying the instance.

core_authentication_service/forms.py
```diff
@@ -0,0 +1,193 @@
...
+
+    @cached_property
+    def get_questions(self):
+        return models.SecurityQuestion.objects.prefetch_related(
+            "questionlanguagetext_set").all()
+
+    def clean(self):
+        # Save some time, if we already have errors, just return without doing
+        # extra work.
+        if all(self.errors):
```
cobusc: This must be `any`, not `all`.

core_authentication_service/forms.py
```diff
@@ -0,0 +1,193 @@
...
+
+            # Ensure unique questions are used.
+            question = form.cleaned_data.get("question", None)
+            if question in questions and question is not None:
+                raise forms.ValidationError(
+                    "Each question can only be picked once."
+                )
+            questions.append(question)
+
+        # If all questions are not all None but less than all is selected,
```
cobusc: Maybe: "If not all questions are selected (i.e. has a value other than None), but some have been, raise an error."
It reads nicely with `if not all(questions) and any(questions):`

core_authentication_service/forms.py
```diff
@@ -0,0 +1,193 @@
...
+
+    def __init__(self, questions, language, *args, **kwargs):
+        super(SecurityQuestionForm, self).__init__(*args, **kwargs)
+        self.fields["question"].queryset = questions
+
+        # Choice tuple can't be directly updated. Update only the widget choice
+        # text, value is used for validation and saving.
+        updated_choices = []
+        for choice in self.fields["question"].widget.choices:
+            choice = list(choice)
```
cobusc: `choice` is a tuple, right? So there should be no need to convert to a list. You can still use indexing, e.g. `choice[0]`.

core_authentication_service/forms.py
AltusBarry: The default password middlewares add help text that is not accurate when not on high security, as the specific validation gets short circuited later in the clean method..

core_authentication_service/models.py
```diff
@@ -0,0 +1,71 @@
...
+from django.conf import settings
+from django.contrib.auth.models import AbstractUser
+from django.core.exceptions import ValidationError
+from django.db import models
+
+
+# NOTE: Changing AUTH_USER_MODEL will cause migration 0001 from otp_totp to
+# break once migrations have already been run once.
+class CoreUser(AbstractUser):
+    id = models.UUIDField(primary_key=True, default=uuid.uuid4, editable=False)
```
cobusc: Please use version 1 UUID. It includes a timestamp, so it is an incrementing value which plays nicely with databases. Totally random primary keys have an unnecessary overhead.

core_authentication_service/models.py
```diff
@@ -0,0 +1,71 @@
...
+    email_verified = models.BooleanField(default=False)
+    nickname = models.CharField(blank=True, null=True, max_length=30)
+    msisdn = models.CharField(blank=True, null=True, max_length=16)
+    msisdn_verified = models.BooleanField(default=False)
+    gender = models.IntegerField(blank=True, null=True)
```
cobusc: Should this not be a `ChoiceField`?

core_authentication_service/models.py
```diff
@@ -0,0 +1,71 @@
...
+    birth_date = models.DateTimeField(blank=True, null=True)
```
cobusc: This should be a `DateField`, not `DateTimeField`.

core_authentication_service/models.py
```diff
@@ -0,0 +1,71 @@
...
+        verbose_name_plural = "Countries"
+
+    def __str__(self):
+        return "%s - %s" % (self.code, self.name)
+
+
+class UserSecurityQuestion(models.Model):
+    user = models.ForeignKey("CoreUser")
+    answer = models.TextField()
+    language_code = models.CharField(max_length=3, choices=settings.LANGUAGES)
```
cobusc: Can we set the `min_length` to 3 as well?

core_authentication_service/models.py
```diff
@@ -0,0 +1,71 @@
...
+    question = models.ForeignKey("SecurityQuestion")
+
+    def __str__(self):
+        return "%s - %s" % (self.language_code, self.question.id)
+    # TODO Hash answer in save()
```
cobusc: This makes sense. We should sanitise, lowercase and then hash.

core_authentication_service/models.py
```diff
@@ -0,0 +1,71 @@
...
+    gender = models.IntegerField(blank=True, null=True)
+    birth_date = models.DateTimeField(blank=True, null=True)
+    country = models.ForeignKey("Country", blank=True, null=True)
+    avatar = models.ImageField(blank=True, null=True)
+    is_employee = models.BooleanField(default=False)
+    is_system_user = models.BooleanField(default=False)
+    updated_at = models.DateTimeField(auto_now=True)
+
+
+class Country(models.Model):
```
cobusc: The prototype repo contains a management command to load country information which you can bring over when you have time.

core_authentication_service/forms.py
AltusBarry: The reason I am not keen to just raise an error is, it will most likely later on not be an error on our side. But if we just raise, it will look like our app is failing, from a user perspective. 

Personally I feel that it is not directly an issue with our form and the user should not be effected, but we do need some form of logging to ensure we can point out what the issue was.

core_authentication_service/oidc_provider_settings.py
```diff
@@ -1,16 +1,68 @@
...
+    "given_name": lambda user: user.first_name,
+    "locale": lambda user: user.country.code if
+        user.country else None,
+    "name": lambda user: "%s %s" % (user.first_name, user.last_name),
+    "nickname": lambda user: user.username,
+    "phone_number": lambda user: user.msisdn,
+    "phone_number_verified": lambda user: user.msisdn_verified if
+        user.msisdn else None,
+    "picture": lambda user: user.avatar,
+    "preferred_username": lambda user: None,
```
cobusc: I believe this can map to the `alias`.

core_authentication_service/password_validation.py
```diff
@@ -0,0 +1,32 @@
...
+from django.core.exceptions import ValidationError
+
+
+class DiversityValidator(object):
+    """
+    Validate whether the password has uppercase, lowercase, digits and special characters.
+    """
+
+    def validate(self, password, user=None):
+        valid = all([
```
cobusc: Very nice. A slightly variation purely for informative purposes:
```python
CHARSETS = [set(string.ascii_lowercase), set(string.ascii_uppercase), set(string.digits), set(string.punctuation)]
...
password_chars = set(password)
# Check that the password characters comes from all CHARSETS
valid = all(password_chars.intersection(charset) for charset in charsets)
```

core_authentication_service/tests/settings/111.py
```diff
@@ -2,15 +2,28 @@
 from project.settings_base import *
 
 
+SECRET_KEY = os.environ.get("SECRET_KEY", "_n(_w(3!i4-p((jz8(o0fb*_r5fb5t!qh1g@m9%4vryx5lale=")
+
+AUTH_USER_MODEL = "core_authentication_service.CoreUser"
+
+AUTH_PASSWORD_VALIDATORS = AUTH_PASSWORD_VALIDATORS + [
```
cobusc: You can use `+=` here

core_authentication_service/views.py
```diff
@@ -0,0 +1,134 @@
...
+        url = settings.LOGIN_URL
+        if self.security == "high" or self.request.GET.get(
+                "show2fa") == "true":
+            url = reverse("two_factor_auth:setup")
+        elif self.redirect_url:
+            url = self.redirect_url
+        return url
+
+
+# Class based view for more flwxibility if needed.
```
cobusc: I'm still not sure where this would be useful. Maybe add docstring that explains what this view does, i.e. interrogate the cookie and redirect to a URL if found, otherwise defaulting to the LOGIN_URL.
It is also not expected that a `logout` is performed, so mention/motivate that explicitly.
Were going to generate documentation, so having nice docstrings helps on many levels.

core_authentication_service/forms.py
cobusc: Kewl. Please put this in a comment in the code.

core_authentication_service/forms.py
cobusc: You are correct. It is not a programmer error because the values are passed in the URL. I missed that bit.

core_authentication_service/forms.py
AltusBarry: I added a comment that will be pushed with the other PR changes

core_authentication_service/forms.py
```diff
@@ -0,0 +1,193 @@
...
+class RegistrationForm(UserCreationForm):
+
+    class Meta:
+        model = get_user_model()
+        fields = [
+            "username", "first_name", "last_name", "email",
+            "nickname", "msisdn", "gender", "birth_date", "country", "avatar"
+        ]
+
+    def __init__(self, security=None, required=[],*args, **kwargs):
```
cobusc: Almost missed this. One should not use mutable default arguments: `required=[]`. Rather do `required=None` and then in the function `required = required or []`.

core_authentication_service/utils.py
```diff
@@ -0,0 +1,51 @@
+def update_form_fields(form, required=[], validators={}, fields_data={}):
```
cobusc: Do not use mutable default argument: `required=[]`. See comment earlier..

core_authentication_service/forms.py
AltusBarry: Will change, was paranoia more than anything else.

core_authentication_service/forms.py
AltusBarry: No idea, must have been an older implementation I forgot to remove parts of. Will remove.

core_authentication_service/forms.py
AltusBarry: Removed, neither implementation was serving the intended purpose.

core_authentication_service/forms.py
AltusBarry: It is a tuple. I, however, am not sure how to actually update values inside of tuples without converting to a list first.

AltusBarry: And then 30 seconds passes and you realise how. Will change. :woman_facepalming:

core_authentication_service/models.py
AltusBarry: It was indicated as an Integer field on the ERD diagram. I did ask around and no one had explicit problems about an integer field, apart from mapping to genders.

core_authentication_service/models.py
AltusBarry: Actually that needs to go up, some of the django languages have keys exceeding 7.

AltusBarry: ('zh-hans', 'Simplified Chinese'),
 ('zh-hant', 'Traditional Chinese')

AltusBarry: The lowest is 2 chars.

AltusBarry: As for min_length, I do not believe that can be enforced on model level.

core_authentication_service/tests/settings/111.py
AltusBarry: Habit in settings, sometimes order matters and its a little quicker to just toss a extra + and list in front of the original.
Will change.

core_authentication_service/password_validation.py
AltusBarry: Like this more, will implement like this.

core_authentication_service/models.py
AltusBarry: Will port that over at a later date, as we do have some tickets for seed data, which that might be able to fall under. Thanks

core_authentication_service/oidc_provider_settings.py
AltusBarry: In the event I have come across this in the oidc docs somewhere, I probably completely forgot what it is. But I can't seem to find mention of an alias.

core_authentication_service/models.py
AltusBarry: Will ask about this again later. Easy change regardless.

core_authentication_service/oidc_provider_settings.py
AltusBarry: Whoops, didn't refresh, it no longer refers to alias.

AltusBarry: Mapped to username, which is used for nickname.

core_authentication_service/views.py
AltusBarry: And you didn't even mention the typo :smile:

core_authentication_service/forms.py
AltusBarry: I'm not 100% sure I understand why this is an issue? Could you perhaps give me some examples?

AltusBarry: I think I get it has something to do with function definition and python's innate ability to still catch me out with the way it references objects.

AltusBarry: Just late day brain. Get it now.

core_authentication_service/models.py
AltusBarry: TODO, will need a quick chat about sanitise, how it might effect comparisons later on, and if we need to salt it as well.

something like `signing.dumps(answer.lower(), salt="core-auth-answer")`

core_authentication_service/forms.py
```diff
@@ -0,0 +1,190 @@
...
+
+LOGGER = logging.getLogger(__name__)
+
+
+REQUIREMENT_DEFINITION = {
+    "names": ["username", "first_name", "last_name", "nickname"],
+    "picture": ["avatar"]
+}
+
+SECURITY_QUESTION_COUNT = 2
```
cobusc: Can we move these constants to a `constants.py` file please. It may not be obvious that it is defined in `forms.py`.

core_authentication_service/models.py
```diff
@@ -0,0 +1,89 @@
...
+
+from django.conf import settings
+from django.contrib.auth import hashers
+from django.contrib.auth.models import AbstractUser
+from django.core.exceptions import ValidationError
+from django.db import models
+
+
+GENDER_CHOICES = (
+    ("female", "Female"),
```
cobusc: The text needs to be translatable.

core_authentication_service/models.py
```diff
@@ -0,0 +1,89 @@
...
+    ("male", "Male"),
+    ("other", "Other")
+)
+
+
+# NOTE: Changing AUTH_USER_MODEL will cause migration 0001 from otp_totp to
+# break once migrations have already been run once.
+class CoreUser(AbstractUser):
+    id = models.UUIDField(primary_key=True, default=uuid.uuid1, editable=False)
```
cobusc: @AltusBarry can we add the following to the model as well, please:
```
# The email field is not required, but needs to be unique if provided
email = EmailField(_('email address'), blank=True, null=True, unique=True)
```

core_authentication_service/models.py
```diff
@@ -0,0 +1,89 @@
...
+    language_code = models.CharField(max_length=7, choices=settings.LANGUAGES)
+    question = models.ForeignKey("SecurityQuestion")
+
+    # NOTE as always, be aware certain update, create and save paths will never
+    # trigger save() or the post/pre save signals.
+    def save(self, *args, **kwargs):
+        # Make use of django built in password hasher, gives us
+        # "check_password" method for comparison later on. In short, salts and
+        # hashes the text.
+        self.answer = hashers.make_password(self.answer.strip().lower())
```
cobusc: Very nice.

cobusc:
Apologies for the late requests. This is a :+1: after those changes are implemented.

PR #9, opened 2018-01-29
Log a user in after registration if the user needs to complete 2FA setup

cobusc:


authentication_service/views.py
```diff
@@ -137,6 +137,14 @@ def form_valid(self, form):
         # Let the user model save.
         response = super(RegistrationView, self).form_valid(form)
 
+        # When we need to show the option to enable 2FA the newly created
+        # user must be logged in.
+        if self.security == "high" or self.request.GET.get("show2fa") == "true":
+            new_user = authenticate(username=form.cleaned_data['username'],
+                                    password=form.cleaned_data['password1'],
+                                    )
```
ruands: I may be wrong, but this indentation looks off.

ruands:
+1

authentication_service/views.py
cobusc: One off, thanks

cobusc: Oddly enough, PyCharm is suggesting this indentation.

PR #10, opened 2018-01-29
Feature/ge 684

AltusBarry:


authentication_service/management/commands/demo_content.py
```diff
@@ -0,0 +1,37 @@
+import os
```
cobusc: Add comment that this management command is only intended for use when setting up a demo environment.

authentication_service/management/commands/demo_content.py
```diff
@@ -0,0 +1,37 @@
...
+
+from django.contrib.auth.models import User
+from django.core.management.base import BaseCommand, CommandError
+
+from oidc_provider.models import Client
+
+
+class Command(BaseCommand):
```
cobusc: Add `help = 'Setup used for demonstration purposes only'` or something to that effect.

scripts/waitFor.sh
```diff
@@ -1,25 +1,8 @@
...
-# Basic poller to check if the postgres service is available. Has dependency on
-# psql being available.
-
-# At this stage it uses the Docker docs example script:
-# https://docs.docker.com/compose/startup-order/
-#until psql -h "$DB_HOST" -U "postgres" -c '\q'; do
-#    >&2 echo "Postgres is unavailable - sleeping"
-#    sleep 1
-#done
+until nc -z ${DB_HOST:=db} ${DB_PORT:=5432} > /dev/null; do
```
cobusc: I cannot see where `db` is defined. Is there a docker compose service called `db` which it defaults to?

AltusBarry: Docker compose service name can be used as domain name. In this case, db is the db service setup in compose.

PR #12, opened 2018-01-31
feature/GE-715: Reordered claims, for human comparison to oidc docs, …

AltusBarry:
…removed super user only login, changed demo clients

### DIFF
--- a/authentication_service/oidc_provider_settings.py
+++ b/authentication_service/oidc_provider_settings.py
@@ -12,25 +12,26 @@ USER_MODEL = get_user_model()
 # Claims for which the resulting function returns None will be automatically
 # omitted from the response.
 CLAIMS_MAP = {
-    "name": lambda user: "%s %s" % (user.first_name, user.last_name),
-    "given_name": lambda user: user.first_name,
-    "family_name": lambda user: user.last_name,
+    "name": lambda user: "%s %s" % (user.first_name, user.last_name) \
+        if user.first_name and user.last_name else None,
+    "given_name": lambda user: user.first_name if user.first_name else None,
+    "family_name": lambda user: user.last_name if user.last_name else None,
     "middle_name": None,
-    "nickname": lambda user: user.nickname,
+    "nickname": lambda user: user.nickname if user.nickname else None,
     "profile": lambda user: None,
     "preferred_username": lambda user: user.username,
-    "picture": lambda user: user.avatar,
+    "picture": lambda user: user.avatar if user.avatar else None,
     "website": lambda user: None,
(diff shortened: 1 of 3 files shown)


authentication_service/oidc_provider_settings.py
```diff
@@ -12,27 +12,28 @@
 # Claims for which the resulting function returns None will be automatically
 # omitted from the response.
 CLAIMS_MAP = {
-    "address": None,
-    "birthdate": lambda user: user.birth_date,
-    "email": lambda user: user.email,
-    "email_verified": lambda user: user.email_verified if
-        user.email else None,
+    "name": lambda user: "%s %s" % (user.first_name, user.last_name),
```
cobusc: These fields are optional, so we have to check if it is set otherwise we and up with garbage being returned,

authentication_service/oidc_provider_settings.py
```diff
@@ -12,27 +12,28 @@
...
+    "given_name": lambda user: user.first_name,
```
cobusc: `lambda user: user.first_name or None`
This means that the name is filled in if it is not None or blank, otherwise None is returned in which case it will not be included in the response.

authentication_service/oidc_provider_settings.py
```diff
@@ -12,27 +12,28 @@
...
     "family_name": lambda user: user.last_name,
```
cobusc: Same as for first_name

authentication_service/oidc_provider_settings.py
```diff
@@ -12,27 +12,28 @@
...
     "gender": lambda user: user.gender,
-    "given_name": lambda user: user.first_name,
+    "birthdate": lambda user: user.birth_date,
+    "zoneinfo": lambda user: None,
     "locale": lambda user: user.country.code if
         user.country else None,
-    "name": lambda user: "%s %s" % (user.first_name, user.last_name),
-    "nickname": lambda user: user.username,
+    "updated_at": lambda user: user.updated_at,
+    "email": lambda user: user.email,
```
cobusc: Email is optional (and can be blank), so: `lambda user: user.email or None`

authentication_service/views.py
```diff
@@ -44,23 +44,6 @@ def __init__(self, **kwargs):
         pipe.delete(get_username_blocked_cache_key(None))
         pipe.execute()
 
-    def done(self, form_list, **kwargs):
```
cobusc: Thanks for removing this.

cobusc:
:+1:

PR #13, opened 2018-02-01
feature/flush-domain-redirect-sessions: Added middleware

AltusBarry:


authentication_service/middleware.py
```diff
@@ -0,0 +1,14 @@
+from urllib.parse import urlparse
+
+from django.utils.deprecation import MiddlewareMixin
+
+
+class OIDCSessionManagementMiddleware(MiddlewareMixin):
+    def process_response(self, request, response):
+        if response.status_code == 302:
+            current_host = request.get_host()
```
cobusc: This is never used as far as I can see

authentication_service/middleware.py
```diff
@@ -0,0 +1,14 @@
...
+            location = response.get("Location", "")
+            parsed_url = urlparse(location)
+            if parsed_url.netloc != "" and location != parsed_url.netloc:
```
cobusc: Was the intention to use the host here?

cobusc:
:+1:

PR #14, opened 2018-02-01
Feature/ge 747

ruands:
Added profile edit pages and tests.


authentication_service/forms.py
```diff
@@ -188,3 +189,38 @@ def __init__(self, questions, language, *args, **kwargs):
...
+        model = get_user_model()
+        fields = [
+            "first_name", "last_name", "nickname", "email", "msisdn", "gender",
+            "birth_date", "country", "avatar"
+        ]
+
+    def __init__(self, *args, **kwargs):
+        super(EditProfileForm, self).__init__(*args, **kwargs)
+
+        # Remove either msisdn or email
```
cobusc: The rule here is that you must have an msisdn or an email or both. You can remove the removal code :)

authentication_service/forms.py
```diff
@@ -188,3 +189,38 @@ def __init__(self, questions, language, *args, **kwargs):
...
+        if self.instance.email:
+            self.fields.pop("msisdn")
+        if self.instance.msisdn:
+            self.fields.pop("email")
+
+    # Make field required
+    email = forms.EmailField(
```
cobusc: Why is email made required here?

authentication_service/models.py
```diff
@@ -34,6 +34,17 @@ class CoreUser(AbstractUser):
     is_system_user = models.BooleanField(default=False)
     updated_at = models.DateTimeField(auto_now=True)
 
+    def save(self, *args, **kwargs):
+        # If email or msisdn has changed, their verified flags need
+        # to be updated.
+        original = CoreUser.objects.filter(pk=self.pk).first()
```
cobusc: This method does a database lookup every time `save()` is called.
I think it can be avoided.
Look at something like this: https://stackoverflow.com/questions/1355150/django-when-saving-how-can-you-check-if-a-field-has-changed?answertab=votes#tab-top
* Make a copy of the fields to watch in `__init__()`
* Check against the copies when doing `save()`
* Update the copies after saving.
@AltusBarry your thoughts?

authentication_service/views.py
```diff
@@ -44,6 +47,24 @@ def __init__(self, **kwargs):
         pipe.delete(get_username_blocked_cache_key(None))
         pipe.execute()
 
+    def done(self, form_list, **kwargs):
+        if self.get_user().is_superuser:
```
cobusc: I think this needs to be removed. I know I asked for it earlier, but it was removed after that, so I don't think it should be reintroduced again.

At this point all users need to be logged in (otherwise they cannot edit their profiles)

authentication_service/models.py
AltusBarry: I like this solution +1

cobusc: @AltusBarry the implemented solution or the proposed solution? :)

AltusBarry: That would be proposed.

authentication_service/forms.py
ruands: Removed since field can be empty if msisdn is present.

authentication_service/models.py
```diff
@@ -34,6 +34,22 @@ class CoreUser(AbstractUser):
     is_system_user = models.BooleanField(default=False)
     updated_at = models.DateTimeField(auto_now=True)
 
+    def __init__(self, *args, **kwargs):
+        super(CoreUser, self).__init__(*args, **kwargs)
+        self.original_email = self.email
```
AltusBarry: rather use underscore variable names `_original_email`

authentication_service/models.py
```diff
@@ -34,6 +34,22 @@ class CoreUser(AbstractUser):
...
+        self.original_msisdn = self.msisdn
+
+    def save(self, *args, **kwargs):
+        # If email or msisdn has changed, their verified flags need
+        # to be updated.
+        if self.email != self.original_email:
+            self.original_email = self.email
```
AltusBarry: Why assign the new email to the old one's variable?

authentication_service/templates/authentication_service/profile/edit_profile.html
```diff
@@ -0,0 +1,14 @@
+{% extends "base.html" %}
+
+{% block content %}
+    <form method="post">
+        {% csrf_token %}
+        {{ form.as_p }}
+        <button type="submit">Update</button>
+        <a href="{% url "update_password" %}">Password Update</a>
```
AltusBarry: We need trans tags in our templates.

authentication_service/templates/authentication_service/profile/edit_profile.html
```diff
@@ -0,0 +1,14 @@
+{% extends "base.html" %}
+
+{% block content %}
+    <form method="post">
+        {% csrf_token %}
+        {{ form.as_p }}
```
AltusBarry: This pulls in form renderers, use `as_div`

authentication_service/tests/test_forms.py
```diff
@@ -450,3 +452,31 @@ def test_validation(self):
...
+        )
+        cls.user.save()
+
+        cls.user_1 = get_user_model().objects.create_user(
+            username="testuser1",
+            msisdn="+27821234567"
+        )
+        cls.user_1.save()
+
+    def test_default_state(self):
```
AltusBarry: Shouldn't error handling also be tested here?

authentication_service/tests/test_forms.py
```diff
@@ -450,3 +452,31 @@ def test_validation(self):
...
+
+    @classmethod
+    def setUpTestData(cls):
+        cls.user = get_user_model().objects.create_user(
+            username="testuser",
+            email="wrong@email.com"
+        )
+        cls.user.save()
+
+        cls.user_1 = get_user_model().objects.create_user(
```
AltusBarry: This user is never used.

authentication_service/tests/test_models.py
```diff
@@ -41,3 +41,31 @@ def test_answer_hashing(self):
...
+    @classmethod
+    def setUpTestData(cls):
+        super(UserModelTestCase, cls).setUpTestData()
+        cls.user = get_user_model().objects.create(
+            username="username1", email="someverified@email.com"
+        )
+        cls.user.email_verified = True
+        cls.user.save()
+
+    def test_email_verification(self):
```
AltusBarry: Also test if the form explicitly posts the old value. It should stay True.

authentication_service/tests/test_views.py
```diff
@@ -254,3 +257,61 @@ def test_incorrect_field_logger(self):
...
+        cls.twofa_user.save()
+
+        cls.totp_device = TOTPDevice.objects.create(
+            user=cls.twofa_user,
+            name="default",
+            confirmed=True,
+            key=random_hex().decode()
+        )
+
+    def test_profile_edit(self):
```
AltusBarry: Also test if the form explicitly posts the old value. It should stay True.

I know the form always sends it anyway, but rather safe than sorry.

authentication_service/views.py
```diff
@@ -44,6 +47,23 @@ def __init__(self, **kwargs):
         pipe.delete(get_username_blocked_cache_key(None))
         pipe.execute()
 
+    def done(self, form_list, **kwargs):
```
AltusBarry: Isn't this the 2fa code copied anyway?

authentication_service/views.py
```diff
@@ -195,3 +215,89 @@ def dispatch(self, request, *args, **kwargs):
...
         return response
+
+
+class EditProfileView(UpdateView):
+    template_name = "authentication_service/profile/edit_profile.html"
+    form_class = forms.EditProfileForm
+    success_url = None
+
+    def dispatch(self, request, *args, **kwargs):
+        self.success_url = self.request.GET.get("redirect_url")
```
AltusBarry: We should probably make getting these values a dispatch decorator, I see we'll be using it on pretty much all the views.

Probably ticket it as a refactor candidate or refactor now.

authentication_service/views.py
```diff
@@ -195,3 +215,89 @@ def dispatch(self, request, *args, **kwargs):
...
+        return super(EditProfileView, self).dispatch(request, *args, **kwargs)
+
+    def get_context_data(self, **kwargs):
+        context = super(EditProfileView, self).get_context_data(**kwargs)
+        # Check if user has 2fa enabled
+        if default_device(self.request.user):
+            context["2fa_enabled"] = True
+        return context
+
+    def get_template_names(self):
```
AltusBarry: Make a ThemeTemplateMixin and add it to the views that need to swap templates on theme, can probably do the dispatch work too.

authentication_service/views.py
```diff
@@ -195,3 +215,89 @@ def dispatch(self, request, *args, **kwargs):
...
+    success_url = None
+
+    def dispatch(self, request, *args, **kwargs):
+        self.success_url = self.request.GET.get("redirect_url")
+        self.theme = self.request.GET.get("theme")
+        return super(EditProfileView, self).dispatch(request, *args, **kwargs)
+
+    def get_context_data(self, **kwargs):
+        context = super(EditProfileView, self).get_context_data(**kwargs)
+        # Check if user has 2fa enabled
```
AltusBarry: Open line above.

authentication_service/views.py
```diff
@@ -195,3 +215,89 @@ def dispatch(self, request, *args, **kwargs):
...
+        self.theme = self.request.GET.get("theme")
+        return super(UpdatePasswordView, self).dispatch(
+            request, *args, **kwargs)
+
+    def get_object(self, queryset=None):
+        return self.request.user
+
+    def get_form_kwargs(self):
+        kwargs = super(UpdatePasswordView, self).get_form_kwargs()
+        kwargs["user"] = kwargs.pop("instance")
```
AltusBarry: THe user kwarg I assume is `PasswordChangeForm` specific?

Why not use rather  `self.get_object()`? Is there an event where the instance gets updated and the user model not? Or does the session user not get updated after instance save?

authentication_service/views.py
```diff
@@ -195,3 +215,89 @@ def dispatch(self, request, *args, **kwargs):
...
+        template_names = []
+
+        if self.template_name is not None:
+            template_names = [self.template_name]
+
+        # Any extra templates need to be before the base.
+        template_names = [
+            "authentication_service/profile/update_password_%s.html" % self.theme,
+        ] + template_names
+        return template_names
```
AltusBarry: Check applicable comments on other view as well. Mixin/decorator related.

authentication_service/views.py
ruands: Using `self.get_object()`

authentication_service/views.py
```diff
@@ -1,23 +1,75 @@
...
+
+    def get_template_names(self):
+        template_names = []
+
+        if self.template_name is not None:
+            template_names = [self.template_name]
+
+        # Any extra templates need to be before the base.
+        template_names = [
+            "%s_%s.html" % (self.TEMPLATE_PREFIX, self.theme),
```
cobusc: Can we make this slightly more elegant? Currently the "default" name will be `_sometheme.html`. Can we change the code that it is `sometheme.html`? Something like
```
"{}{}{}".format(self.TEMPLATE_PREFIX, "_" if self.TEMPLATE_PREFIX else "", self.theme)
```

authentication_service/views.py
```diff
@@ -1,23 +1,75 @@
...
+        ] + template_names
+        return template_names
+
+
+class RedirectMixin:
+    """This mixin gets the redirect url parameter from the request url. This url
```
cobusc: Minor nitpick, but can we use `URL` instead of `url`?

authentication_service/views.py
```diff
@@ -1,23 +1,75 @@
...
+    """
+    success_url = None
+
+    def dispatch(self, *args, **kwargs):
+        self.redirect_url = self.request.GET.get("redirect_url")
+        return super(RedirectMixin, self).dispatch(*args, **kwargs)
+
+    def get_success_url(self):
+        url = settings.LOGIN_URL
+        if hasattr(
```
cobusc: Alternative:
```
if getattr(self, security, "") == "high" or self.request.GET.get("show2fa") == "true":
```

authentication_service/views.py
```diff
@@ -1,23 +1,75 @@
...
+    """This mixin gets the redirect url parameter from the request url. This url
+    is used as the success_url attribute. If no redirect_url is set, it will
+    default to the Login url.
+
+    For registration, this mixin also checks the security level of the request.
+    If the security level is high, the success url will redirect to 2FA setup.
+
+    TODO: Security should be moved out.
+    """
+    success_url = None
```
cobusc: Can we call this the `redirect_url`?

authentication_service/views.py
```diff
@@ -195,3 +224,62 @@ def dispatch(self, request, *args, **kwargs):
         response.delete_cookie(REDIRECT_COOKIE_KEY)
         logout(request)
         return response
+
+
+class EditProfileView(ThemeMixin, RedirectMixin, UpdateView):
+    TEMPLATE_PREFIX = "authentication_service/profile/edit_profile"
+    template_name = "authentication_service/profile/edit_profile.html"
+    form_class = forms.EditProfileForm
+    success_url = None
```
cobusc: `redirect_url` if you're OK with the name change

authentication_service/views.py
```diff
@@ -195,3 +224,62 @@ def dispatch(self, request, *args, **kwargs):
...
+
+    def get_context_data(self, **kwargs):
+        context = super(EditProfileView, self).get_context_data(**kwargs)
+
+        # Check if user has 2fa enabled
+        if default_device(self.request.user):
```
cobusc: Shorthand form:
```
context["2fa_enabled"] = bool(default_device(self.request.user))
```
This means that the context variable is always set explicitly though, which is slightly different logic than what is here.

cobusc:
Lefts some comments/suggestions, but a :+1: overall.

authentication_service/views.py
ruands: Because we override the `get_success_url` method, we don't need to have the attribute `success_url` or `redirect_url`.

cobusc:
:+1:

PR #18, opened 2018-02-06
Cleanup

cobusc:
* Removed deprecated fields from model
* Set database credentials to match the ones used in the Makefile
* Try to infer a nickname if one is not provided.


ruands:
+ 1

PR #22, opened 2018-02-07
Added ability to hide fields or groups of fields on a form

cobusc:


authentication_service/forms.py
```diff
@@ -53,6 +59,11 @@ def __init__(self, security=None, required=None, *args, **kwargs):
             REQUIREMENT_DEFINITION.get(field, [field]) for field in required
         ))
 
+        hidden = hidden or []
```
ruands: Could the default for hidden in the signature be `[]`? I have never tried it.

authentication_service/utils.py
```diff
@@ -16,19 +20,23 @@ def update_form_fields(form, required=None, validators=None, fields_data=None):
             }
         }
 
-    Helper method for setting field and widget atrributes, can
+    Helper method for setting field and widget attributes, can
     be used for any form instance. Sets attributes on both fields and widgets.
     """
     required = required or []
+    hidden = hidden or []
```
ruands: Same here if possible.

ruands:
Just that one question that I have, else +1

authentication_service/forms.py
cobusc: This pattern is to avoid `hidden=[]` in the function signature, because `[]` is a mutable variable that gets created only once and can be modified, with unintended consequences.
PyCharm warns against this too, which is nice.

PR #24, opened 2018-02-09
Add option to provide age instead of birth date when registratering

cobusc:


authentication_service/management/commands/demo_content.py
```diff
@@ -43,14 +43,16 @@ def handle(self, *args, **options):
...
+                                               birth_date=datetime.date(2000, 1, 1))
         user.set_password("local")
         user.save()
 
         # End User
         end_user = get_user_model().objects.create(
             username="enduser", first_name="End", last_name="User",
-            email="enduser@here.com", nickname="l33t"
+            email="enduser@here.com", nickname="l33t",
+            birth_date=datetime.date(2000, 1, 1)
```
ruands: Is this not required for `system_user` as well?

cobusc: It is required for _all_ users.

ruands:
+1 :)

PR #25, opened 2018-02-09
Make birth_date form field not be required (although it is still requ…

cobusc:
…ired on the model). The previous implementation worked fine, but still rendered the field as required (despite the widget attribute being set as not required)


ruands:
+1

ruands:
👍

PR #27, opened 2018-02-12
Added more theme templates for login and edit profile pages

ruands:


authentication_service/views.py
```diff
@@ -25,7 +25,7 @@ class ThemeMixin:
     sets the appropriate template.
     """
     TEMPLATE_PREFIX = ""
-    theme = None
+    # theme = None
```
cobusc: Can this be removed?

cobusc: Or should it stay?

ruands: Oh, yes 👍

cobusc:
:+1: Just that one comment that needs to be resolved, please.

PR #29, opened 2018-02-19
GE-789: Password reset

ruands:
- Added password reset using email/security questions.
- Added tests for password reset.


authentication_service/tasks.py
```diff
@@ -0,0 +1,25 @@
...
+
+logger = logging.getLogger()
+
+
+@task(name="send_email_task")
+def send_email_task(subject_template_name, email_template_name,
+                  context, from_email, to_email, html_email_template_name=None):
+    subject = loader.render_to_string(subject_template_name, context)
+    # Email subject *must not* contain newlines
+    subject = ''.join(subject.splitlines())
```
cobusc: Should this maybe be `' '.join(subject.splitlines())`? (Note the space)

requirements/requirements.txt
```diff
@@ -20,3 +20,6 @@ Pillow==5.0.0
 
 django-defender>=0.5.1,<0.6
 dateutils>=0.6.6,<1.0.0
+
+# Celery
+celery==4.0.0
```
cobusc: `>=4.1.0,<5.0` please

project/settings.py
```diff
@@ -88,6 +88,10 @@
     "enable-bem-classes": True
 }
 
+CELERY_USE_TZ = USE_TZ
+CELERY_BROKER_URL = os.environ.get("CELERY_BROKER_URL", "amqp://myuser:mypassword@localhost:5672//")
+CELERY_RESULT_BACKEND = os.environ.get("CELERY_RESULT_BACKEND", "amqp://myuser:mypassword@localhost:5672//")
```
cobusc: Do we have to have 2 different config entries? Can the one maybe point to the other?

requirements/requirements.txt
cobusc: Also add to test settings, please.

project/settings.py
ruands: I grabbed these from Power of Play since I know they work. I can try get one of them to point to the other.

authentication_service/tasks.py
ruands: This is the exact code from the Django `PasswordResetForm`'s `send_mail` method.

cobusc: My question still stands. "Foo\nBar" will become "FooBar" using this code. Would "Foo Bar" not be better?

ruands: Oh I see. Yes, I'll change it.

cobusc:
:+1:

PR #31, opened 2018-02-20
feature/demo-changes: Added temp json roles file, added 2 demo regist…

AltusBarry:
…er templates, updated claims to return json roles

### DIFF
--- a/authentication_service/demo/roles.json
+++ b/authentication_service/demo/roles.json
@@ -1,0 +1,7 @@
+{
+    "client_id_1": {
+        "": ["Admin", "Editor", "Trainee", "Manager"]
+    },
+    "client_id_2": {
+    }
+}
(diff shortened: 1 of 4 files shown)


authentication_service/demo/roles.json
```diff
@@ -0,0 +1,7 @@
+{
+    "client_id_1": {
+        "": ["Admin", "Editor", "Trainee", "Manager"]
```
cobusc: Blank key?

AltusBarry: Will need to be added locally as users get created during the demo.

cobusc: We have two demo users that gets created: end user and system user IIRC. Maybe add them?

cobusc:
:+1:

authentication_service/demo/roles.json
AltusBarry: Will add them locally, didn't want to introduce extra logic to add them to something we only added for the demo.

PR #34, opened 2018-02-22
feature/GE-827: Removed mixin, added middleware

AltusBarry:
No need for a subclass with mixin anymore. Should allow for theme templates for 3rd party apps to "just work".

### DIFF
--- a/authentication_service/middleware.py
+++ b/authentication_service/middleware.py
@@ -47,7 +47,6 @@ class ThemeManagementMiddleware(MiddlewareMixin):
             response.set_cookie(
                 self.cookie_key, value=theme, httponly=True
             )
-        if theme:
             templates = {"original": [], "new": []}
             for path in response.template_name:
                 file_name = path[path.rindex("/")+1:]
--- a/authentication_service/middleware.py
+++ b/authentication_service/middleware.py
@@ -41,7 +41,8 @@ class ThemeManagementMiddleware(MiddlewareMixin):
     cookie_key = "ge_theme_middleware_cookie"
 
     def process_template_response(self, request, response):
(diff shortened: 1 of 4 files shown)


authentication_service/middleware.py
```diff
@@ -35,3 +35,33 @@ def process_response(self, request, response):
...
+                templates["new"].append({"name": name, "path": dir_path})
+                templates["original"].append(file_name)
+
+            joined_names = ",".join(templates["original"])
+            for template in templates["new"]:
+                prepend_list = []
+                if template["name"] not in joined_names:
+                    prepend_list.append("%s%s" % (
+                        template["path"], template["name"]))
+                response.template_name = prepend_list + response.template_name
```
cobusc: Is `response.template_name` a list?

cobusc:
Awesome rework. +1

authentication_service/middleware.py
AltusBarry: Yip, dunno why the name is singular. Its the culmination of the templates after `get_template_names` has run I believe.

authentication_service/middleware.py
```diff
@@ -35,3 +35,33 @@ def process_response(self, request, response):
...
+    def process_template_response(self, request, response):
+        theme = request.GET.get("theme", None) or request.COOKIES.get(
+            self.cookie_key)
+        if theme:
+            response.set_cookie(
+                self.cookie_key, value=theme, httponly=True
+            )
+            templates = {"original": [], "new": []}
+            for path in response.template_name:
+                file_name = path[path.rindex("/")+1:]
```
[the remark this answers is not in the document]
AltusBarry: Seems like a better idea, will use.

PR #38, opened 2018-02-28
Feature/ge 866

AltusBarry:


authentication_service/templates/authentication_service/email/delete_account.html
```diff
@@ -0,0 +1,5 @@
+The following user would like to delete their Girl Effect account.
+username: {{ user.username }}
```
cobusc: Can we add the  user id (UUID) as well, please?

scripts/startDjango.sh
```diff
@@ -3,4 +3,5 @@
 ADDRESS_VALUE=${ADDRESS:=0.0.0.0:8000}
 
 python manage.py migrate --noinput
+celery -A authentication_service worker -l info &
```
cobusc: Please add TODO to remove this and run it as its own service in the docker-compose environment

cobusc:
:+1:

PR #41, opened 2018-03-05
Removed OIDC-related definitions

cobusc:
It is already implemented and not part of what Swagger provides


ruands:
Looks good @cobusc

cobusc:
For historical purposes: I have left parts of the spec commented out as we may want to re-introduce it at some point for documentation purposes, but we do not need it right now for code generation purposes.

PR #42, opened 2018-03-06
Feature/ge 880

AltusBarry:


authentication_service/forms.py
```diff
@@ -379,3 +381,51 @@ class DeleteAccountForm(forms.Form):
...
+                }
+            }
+            # Update the actual fields and widgets.
+            update_form_fields(
+                self,
+                fields_data=fields_data,
+            )
+
+    def clean_new_password2(self):
+        # Short circuit normal validation if not high security.
```
cobusc: This comment does not make sense. If a user has an organisational unit we _do_ want high security.

authentication_service/forms.py
```diff
@@ -379,3 +381,51 @@ class DeleteAccountForm(forms.Form):
...
+        if self.user.organisational_unit:
```
cobusc: According to the comment this should be: `if not self.organisational_unit`

cobusc:
:+1: when you add a nice docstring to the `SetPasswordForm` class.

cobusc:
:+1:

PR #43, opened 2018-03-06
Feature/ge 858

ruands:


authentication_service/integration.py
```diff
@@ -0,0 +1,125 @@
...
+        return CoreUser.objects.filter(id=user_id).values(*USER_VALUES).get()
+
+    @staticmethod
+    def user_update(request, body, user_id, *args, **kwargs):
+        """
+        :param request: An HttpRequest
+        :param body: dict A dictionary containing the parsed and validated body
+        :param user_id: string A UUID value identifying the user.
+        """
+        instance, created = CoreUser.objects.get_or_create(id=user_id)
```
cobusc: Update should never create a user. It should do a `get_or_404`

authentication_service/integration.py
```diff
@@ -0,0 +1,125 @@
...
+        if not created:
+            for attr, value in body.items():
+                setattr(instance, attr, value)
```
cobusc: Do the `setattr` in a `try...except` block. Log when something goes wrong.

authentication_service/integration.py
```diff
@@ -0,0 +1,125 @@
...
+            instance.save()
+        return CoreUser.objects.filter(id=user_id).values(*USER_VALUES).get()
```
cobusc: Is there not a way to return the instance as a dictionary? Doing another DB lookup feels unnecessary

ruands: The `avatar` field caused an issue, but I have fixed it since.

authentication_service/integration.py
```diff
@@ -0,0 +1,137 @@
...
+    def client_list(request, offset=None, limit=None, client_ids=None,
+                    client_token_id=None, *args, **kwargs):
+        """
+        :param request: An HttpRequest
+        :param offset (optional): integer An optional query parameter specifying the offset in the result set to start from.
+        :param limit (optional): integer An optional query parameter to limit the number of results returned.
+        :param client_ids (optional): string An optional query parameter to filter by a list of client.id.
+        :param clent_token_id (optional): string An optional query parameter to filter by a single client.client_id.
+        """
+        if client_ids:
```
cobusc: The options are not exclusive. All of them can be used at the same time, so you cannot use `if..elif..else`.
Do something like:
```
clients = Client.objects.values(*CLIENT_VALUES)
if client_ids:
    clients = clients.filter(pk__in-==client_ids)

if client_token_id:
    clients = clients.filter(client_id=client_token_id)

start = int(offset) if offset else settings.DEFAULT_LISTING_OFFSET 
limit = int(limit) if limit else settings.DEFAULT_LISTING_LIMIT
clients = clients[start:start+limit]  # Slice before evaluating query
return list(clients)  # Evaluate query when casting to a list
```

authentication_service/integration.py
```diff
@@ -0,0 +1,137 @@
...
+                client for client in Client.objects.filter(
+                    pk__in=client_ids).values(*CLIENT_VALUES)
+            ]
+        elif client_token_id:
+            result = [
+                client for client in Client.objects.filter(
+                    client_id=client_token_id).values(*CLIENT_VALUES)
+            ]
+        else:
+            result = [
```
cobusc: This is a bad way of slicing. It will literally pull all the data (matching the query) from the database and only then apply the offset and limit slicing. See my example where the slicing is done on the QuerySet, which is how it should be done.
https://docs.djangoproject.com/en/2.0/ref/models/querysets/#when-querysets-are-evaluated (see `slicing` and `list` sections)

authentication_service/integration.py
```diff
@@ -0,0 +1,137 @@
...
+            client_id=client_id).values(*CLIENT_VALUES).get()
+        return result
+
+    @staticmethod
+    def user_list(request, offset=None, limit=None, email=None,
+                  username_prefix=None, user_ids=None, *args, **kwargs):
+        """
+        :param request: An HttpRequest
+        """
+        if email:
```
cobusc: These options are also not mutually exclusive. Multiple can be passed. Please change this to look like the example I gave for clients.

authentication_service/integration.py
```diff
@@ -0,0 +1,137 @@
...
+            result = [
+                user for user in CoreUser.objects.filter(
+                    email=email).values(*USER_VALUES)
+            ]
+        elif username_prefix:
+            result = [
+                user for user in CoreUser.objects.filter(
+                    username__contains=username_prefix).values(*USER_VALUES)
```
cobusc: The `username_prefix` is a _prefix_, so you have to check that the username starts with it, i.e. `username__startswith`

authentication_service/integration.py
```diff
@@ -0,0 +1,137 @@
...
+            ]
+        elif user_ids:
+            result = [
+                user for user in CoreUser.objects.filter(
+                    id__in=user_ids).values(*USER_VALUES)
+            ]
+        else:
+            result = [
```
cobusc: See previous comment about why this way of slicing is bad.

authentication_service/integration.py
```diff
@@ -0,0 +1,137 @@
...
+        :param request: An HttpRequest
+        :param body: dict A dictionary containing the parsed and validated body
+        :param user_id: string A UUID value identifying the user.
+        """
+        instance = get_object_or_404(CoreUser, id=user_id)
+        for attr, value in body.items():
+            try:
+                setattr(instance, attr, value)
+            except Exception as e:
+                print(e)
```
cobusc: I know I asked you to stick the `setattr` in a `try..except...` block, but now I realise it may never raise an exception. If that is the case, then you can remove it along with the `print` statement. If it stays, then please change the `print` to a proper log line. e.g.
```
import logging

LOGGER = logging.getLogger(__name__)

LOGGER.error("Failed to set user attribute {}: {}".format(attr, e))
```

authentication_service/integration.py
```diff
@@ -0,0 +1,137 @@
...
+
+        instance.save()
+        obj = {}
```
cobusc: Maybe don't call the dictionary `obj`, but rather `result` or something like that.

authentication_service/tests/settings/111.py
```diff
@@ -85,3 +85,11 @@
     "enable-bem-classes": True
 }
 
+# API Settings
+STUBS_CLASS = "authentication_service.integration.Implementation"
+SWAGGER_API_VALIDATE_RESPONSES = True
+DEFAULT_LISTING_LIMIT = 20
+MAX_LISTING_LIMIT = 100
```
cobusc: Where is this used?

authentication_service/tests/settings/111.py
```diff
@@ -85,3 +85,11 @@
...
+MIN_LISTING_LIMIT = 1
```
cobusc: Where is this used?

authentication_service/tests/test_integration.py
```diff
@@ -0,0 +1,200 @@
...
+        # Test offset
+        response = self.client.get("/api/v1/clients?offset=1")
+        self.assertEqual(len(response.json()), 1)
+        self.assertContains(response, "%s" % self.client_2.id)
+
+        # Test list using client.id
+        response = self.client.get(
+            "/api/v1/clients?client_ids=%s&client_ids=%s" % (
+                self.client_1.id, self.client_2.id)
+        )
```
cobusc: Add a test using a combination of parameters as well, please.

authentication_service/tests/test_integration.py
```diff
@@ -0,0 +1,200 @@
...
+        self.client.login(username="test_user_2", password="password")
+
+        # Test read
+        response = self.client.get(
+            "/api/v1/clients/%s" % self.client_1.client_id)
+        self.assertContains(response, "test_client_id_1")
+
+        # Check that data contains all the client fields per the swagger schema
+        fields = response.json()
+        for field in CLIENT_VALUES:
```
cobusc: Instead of just checking the fields you should simply be able to validate against the schema.
Something like `jsonschema.validate(response.json(), schemas.client)`

authentication_service/tests/test_integration.py
```diff
@@ -0,0 +1,200 @@
...
+        # Authorize user
+        self.client.login(username="test_user_2", password="password")
+
+        # Test read
+        response = self.client.get("/api/v1/users/%s" % self.user_1.id)
+        self.assertContains(response, "test_user_1")
+
+        # Check that data contains all the user fields per the swagger schema
+        fields = response.json()
+        for field in USER_VALUES:
```
cobusc: Instead of just checking the fields you should simply be able to validate against the schema.
Something like `jsonschema.validate(response.json(), schemas.user)`

cobusc:
@ruands I left some comments

authentication_service/tests/settings/111.py
ruands: Forgot to implement.

authentication_service/tests/settings/111.py
ruands: Forgot to implement.

authentication_service/integration.py
```diff
@@ -0,0 +1,138 @@
...
+        """
+        offset = int(offset if offset else settings.DEFAULT_LISTING_OFFSET)
+        limit = set_listing_limit(limit)
+
+        clients = Client.objects.values(*CLIENT_VALUES)
+
+        q_objects = Q()
+
+        if client_ids:
+            q_objects |= Q(id__in=client_ids)
```
cobusc: If I do a query: `foo=1&bar=2` I expect to get results where with `foo=1` AND `bar=2`, not OR. The code as in the example I provided in the previous PR was correct.

authentication_service/integration.py
```diff
@@ -0,0 +1,138 @@
...
+                  username_prefix=None, user_ids=None, *args, **kwargs):
+        """
+        :param request: An HttpRequest
+        """
+        offset = int(offset if offset else settings.DEFAULT_LISTING_OFFSET)
+        limit = set_listing_limit(limit)
+
+        users = CoreUser.objects.values(*USER_VALUES)
+
+        q_objects = Q()
```
cobusc: Same comment as the previous one

authentication_service/utils.py
```diff
@@ -61,3 +62,35 @@ def update_form_fields(form, required=None, hidden=None, validators=None, fields
...
+def strip_empty_optional_fields(object_dict):
+    """ We do not need to add fields that contain None or "" to the response,
+    so we strip those fields out of the response. To do this, we iterate over
+    the fields in the input dictionary and check that the value isn't, what we
+    consider, empty. If a field has a value, add that field and value to the
+    output dictionary.
+    :param object_dict: Input dictionary containing possible empty fields.
+    :return: Output dictionary containing only fields that have values.
+    """
+    result = {}
```
cobusc: `return {k: v for k, v in object_dict.iteritems() if v is not None}` should suffice.

authentication_service/utils.py
```diff
@@ -61,3 +62,31 @@ def update_form_fields(form, required=None, hidden=None, validators=None, fields
...
+
+def set_listing_limit(limit):
+    """ Ensures the limit is within bounds or sets the default limit if no limit
+    was specified.
+    :param limit: Amount of objects to return.
+    :return: Either the minimum, maximum or the default limit.
+    """
+    if limit:
+        limit = int(limit)
+        limit = limit if limit <= settings.MAX_LISTING_LIMIT else \
```
cobusc: When limits are exceeded they should return HttpBadRequest (an HTTP 400 response), please.

authentication_service/integration.py
```diff
@@ -0,0 +1,132 @@
...
+
+        clients = Client.objects.values(*CLIENT_VALUES).order_by("id")
+
+        if client_ids:
+            clients = clients.filter(id__in=client_ids)
+
+        if client_token_id:
+            clients = clients.filter(client_id=client_token_id)
+
+        clients = clients[offset:limit]
```
cobusc: `clients[offset:offset+limit]`

authentication_service/integration.py
```diff
@@ -0,0 +1,132 @@
...
+        users = CoreUser.objects.values(*USER_VALUES).order_by("id")
+
+        if user_ids:
+            users = users.filter(id__in=user_ids)
+        if email:
+            users = users.filter(email=email)
+        if username_prefix:
+            users = users.filter(username__startswith=username_prefix)
+
+        users = users[offset:limit]
```
cobusc: `clients[offset:offset+limit]`

authentication_service/integration.py
```diff
@@ -0,0 +1,132 @@
...
+        clients = Client.objects.values(*CLIENT_VALUES).order_by("id")
+
+        if client_ids:
+            clients = clients.filter(id__in=client_ids)
+
+        if client_token_id:
+            clients = clients.filter(client_id=client_token_id)
+
+        clients = clients[offset:limit]
+        return list(clients)
```
cobusc: `return [strip_empty_optional_fields(client) for client in clients]`

authentication_service/integration.py
```diff
@@ -0,0 +1,132 @@
...
+        if email:
+            users = users.filter(email=email)
+        if username_prefix:
+            users = users.filter(username__startswith=username_prefix)
+
+        users = users[offset:limit]
+        result = []
+        for user in list(users):
+            result.append(strip_empty_optional_fields(user))
+        return result
```
cobusc: return [strip_empty_optional_fields(user) for user in users]

authentication_service/integration.py
```diff
@@ -0,0 +1,132 @@
...
+
+    @staticmethod
+    def user_delete(request, user_id, *args, **kwargs):
+        """
+        :param request: An HttpRequest
+        :param user_id: string A UUID value identifying the user.
+        """
+        user = CoreUser.objects.filter(id=user_id)
```
cobusc: `get_or_404()`

authentication_service/integration.py
```diff
@@ -0,0 +1,132 @@
...
+
+
+    @staticmethod
+    def user_read(request, user_id, *args, **kwargs):
+        """
+        :param request: An HttpRequest
+        :param user_id: string A UUID value identifying the user.
+        """
+        result = strip_empty_optional_fields(
+            CoreUser.objects.filter(id=user_id).values(*USER_VALUES).get())
```
cobusc: What happens when the user does not exist?
Also, try something like this:
```
try:
    user = CoreUser.objects.values(*USER_VALUES).get(id=user_id)
    return strip_empty_optional_fields(user)
except CoreUser.DoesNotExist:
   raise Http404()

File: authentication_service/integration.py [hunk 26]
```diff
@@ -0,0 +1,123 @@
...
+
+        if client_token_id:
+            clients = clients.filter(client_id=client_token_id)
+
+        clients = clients[offset:offset + limit]
+        return [strip_empty_optional_fields(client) for client in clients]
+
+
+    @staticmethod
+    def client_read(request, client_id, *args, **kwargs):
```
cobusc: Please implement this similar to the `user_read` function

authentication_service/tests/test_integration.py
```diff
@@ -0,0 +1,221 @@
...
+        # Authorize user
+        self.client.login(username="test_user_2", password="password")
+
+        # Test read
+        response = self.client.get(
+            "/api/v1/clients/%s" % self.client_1.client_id)
+        self.assertContains(response, "test_client_id_1")
+
+        # Validate returned data
+        jsonschema.validate(response.json(), schema=schemas.client)
```
cobusc: This is a general comment: Please add tests for read/delete/updates of ids that does not exist, which should return an HTTP 404. This goes for users and clients.

authentication_service/integration.py
```diff
@@ -0,0 +1,128 @@
...
+
+        clients = Client.objects.values(*CLIENT_VALUES).order_by("id")
+
+        if client_ids:
+            clients = clients.filter(id__in=client_ids)
+
+        if client_token_id:
+            clients = clients.filter(client_id=client_token_id)
+
+        if not clients:
```
cobusc: Noooo

authentication_service/integration.py
```diff
@@ -0,0 +1,128 @@
...
+        users = CoreUser.objects.values(*USER_VALUES).order_by("id")
+
+        if user_ids:
+            users = users.filter(id__in=user_ids)
+        if email:
+            users = users.filter(email=email)
+        if username_prefix:
+            users = users.filter(username__startswith=username_prefix)
+
+        if not users:
```
cobusc: Nooo

cobusc:
:+1:

PR #46, opened 2018-03-15
Feature/ge 813

AltusBarry:


authentication_service/context_processors.py
```diff
@@ -0,0 +1,15 @@
...
+    session_client_name = request.session.get(
+        EXTRA_SESSION_KEY,
+        {}
+    ).get(COOKIES["redirect_client_name"])
+    return {
+        "ge_global_redirect_uri": request.COOKIES.get(
+            COOKIES["redirect_cookie"]
+        ),
+        "ge_global_client_name": request.COOKIES.get(
+            COOKIES["redirect_client_name"], session_client_name
```
cobusc: Is this the client name or client id?

AltusBarry: Name. Its to be displayed on a template.

authentication_service/middleware.py
```diff
@@ -2,8 +2,19 @@
 import logging
 import os
 
+from oidc_provider.lib.endpoints.authorize import AuthorizeEndpoint
+from oidc_provider.lib.errors import (
```
ruands: Didn't know you could import like this. Pretty neat.

authentication_service/context_processors.py
cobusc: If that is the case then this cookie is delicious.

cobusc:
:+1:

PR #49, opened 2018-03-23
Feature/ge 886 and ge 901 theming

markupguy:
@AltusBarry Please check with an eagle eye :) Template names and extends mostly, just incase I made a spelling mistake. I could test most of them, but not all of them as I couldn't find some pages.


AltusBarry:
+1. Million thanks, this looks like it was a pain. Apologies that we didn't have an easier solution in place for you yet. :bowing_woman:

PR #54, opened 2018-03-29
Feature/some theme fun

AltusBarry:


authentication_service/middleware.py
```diff
@@ -48,20 +48,34 @@ def process_response(self, request, response):
...
+
         return response
 
+def fetch_theme(request, key=None):
+    theme = request.GET.get("theme", None) or request.COOKIES.get(key)
+
+    # Next querystring contain an entire url including querystrings. Django
+    # request GET is not aware of the inner querystrings.
+    next_query = request.GET.get("next")
+    if next_query and "theme" in next_query:
```
cobusc: I would parse first and them check if theme is one of the keys in the dict.

authentication_service/middleware.py
```diff
@@ -48,20 +48,34 @@ def process_response(self, request, response):
...
+                response.delete_cookie(COOKIES["ge_theme_middleware_cookie"])
+
         return response
 
+def fetch_theme(request, key=None):
+    theme = request.GET.get("theme", None) or request.COOKIES.get(key)
+
+    # Next querystring contain an entire url including querystrings. Django
+    # request GET is not aware of the inner querystrings.
+    next_query = request.GET.get("next")
```
cobusc: This would override the theme that is explicitly provided in the request. Should this not rather be treated as a fallback?

authentication_service/middleware.py
```diff
@@ -48,20 +48,34 @@ def process_response(self, request, response):
                         current_host, parsed_url.netloc
                     )
                 )
+
+                # Clear theme cookie
+                response.delete_cookie(COOKIES["ge_theme_middleware_cookie"])
+
         return response
 
+def fetch_theme(request, key=None):
```
cobusc: Please add a docstring explaining the steps (priority) in determining the theme to use. We have 3 potential sources:  "theme" parameter, "cookie" and "next" parameter. It must be clear how these are interrogated to avoid confusion.

authentication_service/middleware.py
cobusc: Then why check next if theme is already set?

cobusc: Example of what I had in mind:
```
    # First check if a theme was provided in the URL, or previously set in a cookie
    theme = request.GET.get("theme") or request.COOKIES.get(key)

    # If neither of these are there, we check if there is a `next` URL specified with a theme argument in it.
￼    if theme is None:
￼        next_query = request.GET.get("next")
        if next_query:
            next_query_args = parse_qs(urlparse(next_query).query)
            theme = next_query_args.get("theme")

    return theme
```

cobusc:
:+1:

PR #55, opened 2018-04-04
Feature/ge 924

AltusBarry:


authentication_service/tests/test_forms.py
```diff
@@ -15,6 +17,16 @@
 
 class TestRegistrationForm(TestCase):
 
+    def setUp(self):
+        # Disable setting field hiding, to tests original form behaviour.
+        self.orig_settings = copy.deepcopy(settings.HIDE_FIELDS)
```
cobusc: Instead of doing this by hand, you can decorate the `TestRegistrationForm` class with something like this:
```
@patch.dict("settings.HIDE_FIELDS", {"global_enable": True})
```
It will set the values and restore the original values after the test(s) have executed. It can also be used as a function decorator or context manager.

AltusBarry: I tried that, it didn't seem to patch the correct instance of the django.conf settings. So I fell back to doing it in setUp and tearDown.

AltusBarry: And django's own `override_settings` is a method decorator.

AltusBarry: I did forget about `override_settings` however.

cobusc: I was just going to say that myself :)

PR #56, opened 2018-04-04
Added custom error pages

ruands:


authentication_service/templates/500.html
```diff
@@ -0,0 +1 @@
+Oops! Something went terribly wrong.
```
AltusBarry:
```
<html>
    <body>
        <p>Oops! Something went terribly wrong.</p>
    </body>
</html>
```

authentication_service/templates/404.html
```diff
@@ -0,0 +1,5 @@
+{% extends "base.html" %}
```
AltusBarry: Add the ninyampinga and springster variations as well.

authentication_service/templates/ninyampinga/404.html
```diff
@@ -0,0 +1,12 @@
+{% extends "base.html" %}
```
AltusBarry:
```{% extends "base.html" %}

{% block extrastyle-basic %}<link rel="stylesheet" type="text/css" href="{% static "authentication_service/css/theme/ninyampinga.basic.css" %}" />{% endblock %}
 
 {% block extrastyle-enhanced %}<link rel="stylesheet" type="text/css" href="{% static "authentication_service/css/theme/ninyampinga.enhanced.css" %}" />{% endblock %}
 
 {% block branding %}<img width="70px" src="http://ninyampinga.com/static/img/logos/logo_rwanda.a61da5b16fba.png" class="Header-branding-image">{% endblock %}
```
We need to avoid the situation where we need to keep copy up to date in multiple templates.

authentication_service/templates/ninyampinga/500.html
```diff
@@ -0,0 +1,10 @@
+<!DOCTYPE html>
```
AltusBarry: Check if django middleware kicks in for default 500s, if not these styled ones will never get served. Even then 503s won't be able to pick a specific styled one.

authentication_service/templates/ninyampinga/404.html
ruands: See what you mean 👍

AltusBarry:
+1

PR #61, opened 2018-04-10
OIDC URLs should not form part of the i18n_patterns

cobusc:


ruands:
+1, thanks for this. I missed it.

PR #63, opened 2018-04-13
GE-1006: Update fields

ruands:


authentication_service/forms.py
```diff
@@ -327,15 +339,27 @@ def __init__(self, *args, **kwargs):
...
         hidden_fields = []
+
         # Final overrides from settings
         if settings.HIDE_FIELDS["global_enable"]:
             for field in settings.HIDE_FIELDS["global_fields"]:
                 self.fields[field].required = False
                 self.fields[field].widget.is_required = False
                 hidden_fields.append(field)
 
+        # Since email is required on registration for a system user,
```
AltusBarry: Comment not updated.

AltusBarry:
+1 update comment, please. What a pain this is gonna be to trouble shoot when the requires querystrings makes it seem like things "broke".

PR #65, opened 2018-04-17
Translations for German and Afrikaans

ruands:
- Added German and Afrikaans translations for most of the auth service (should be enough to demo how translations work).

screen shot 2018-04-17 at 6 44 12 pm
screen shot 2018-04-17 at 6 44 26 pm
screen shot 2018-04-17 at 6 47 07 pm

- Attempted to solve the missing .mo files on QA. `Dockerfile` would be ideal, but the environment variables aren't present when the image builds. This can be solved by adding the `ENV` command, but that would set the a variable even when we don't want it so I don't believe this is the right option.
- Added the required `gettext` libs for translations to the `pip install` line in the `Dockerfile`.
- Updated the `startDjango` script to at least allow compilation of translations in docker-compose environments.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -17,4 +17,6 @@ RUN pip3 install --no-cache-dir -r /app/requirements/requirements.txt --src /usr
 
 COPY . /app/
 
+RUN django-admin compilemessages
+
 EXPOSE 8000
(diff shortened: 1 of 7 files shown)


cobusc:
The screenshots look great!

cobusc:
:+1: Thanks

PR #67, opened 2018-04-18
ruands profile fields

ruands:
- Added age field that grabs the users `birth_date` and converts to age.
- When user updates their age, it will convert back into the current day and month and the correct year.

### DIFF
--- a/authentication_service/tests/test_views.py
+++ b/authentication_service/tests/test_views.py
@@ -680,7 +680,7 @@ class DeleteAccountTestCase(TestCase):
             "<input name=\"confirmed_deletion\" type=\"submit\" value=\"Yes\" class=\"Button\" />"
         )
         self.assertContains(response,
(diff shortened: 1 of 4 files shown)


authentication_service/tests/test_views.py
```diff
@@ -679,11 +679,9 @@ def test_view_html_toggle(self):
             response,
             "<input name=\"confirmed_deletion\" type=\"submit\" value=\"Yes\" class=\"Button\" />"
         )
-
-        # TODO re-add later.
-        #self.assertContains(response,
-        #    '<textarea name="reason" cols="40" rows="10" id="id_reason" class=" Textarea "></textarea>'
-        #)
+        self.assertContains(response,
+            "<textarea name=\"reason\" cols=\"40\" rows=\"10\" id=\"id_reason\" class=\" Textarea \">"
```
AltusBarry: Thanks a mil for this.

AltusBarry:
+1

PR #72, opened 2018-04-24
Feature/ge 1029

AltusBarry:


authentication_service/models.py
```diff
@@ -14,28 +17,75 @@
...
+    Custom field to populate concatenated query field.
+
+    Should work for more scenarios than just save() or create()
+    """
+    def pre_save(self, model_instance, add):
+        value = ""
+        fields = [
+            "email", "first_name", "last_name", "msisdn", "nickname", "username"
+        ]
+        for field in fields:
```
cobusc: `value = " ".join(getattr(model_instance, field, " ") for field in fields)` should suffice?

authentication_service/models.py
```diff
@@ -14,28 +17,75 @@
...
 
+class AutoQueryField(models.TextField):
+    """
+    Custom field to populate concatenated query field.
+
+    Should work for more scenarios than just save() or create()
+    """
+    def pre_save(self, model_instance, add):
+        value = ""
+        fields = [
```
cobusc: It would be nice if the fields can be passed to the constructor, making this class reusable.

authentication_service/tests/models.py
```diff
@@ -0,0 +1,5 @@
+from django.db import models
+
+
+class DummyModel(models.Model):
```
cobusc: Where is this used?

authentication_service/models.py
AltusBarry: Did it the long way round, because getattr can return a None and will cause a NoneType to str error.

authentication_service/tests/models.py
AltusBarry: Its in tests, only there to be able to generate test migrations, required to create the pg_tgrm extension in the test db.

AltusBarry: Short answer, nowhere.

authentication_service/models.py
cobusc: Ah, so you mean it should be something like:
`value = " ".join(getattr(model_instance, field, "") or "" for field in fields)`

cobusc:
:+1: Just look at my comments and decide which ones you need to act on.

authentication_service/models.py
AltusBarry: will add that.

PR #74, opened 2018-04-24
New search filters

cobusc:
As described here: https://praekelt.atlassian.net/wiki/spaces/GEM/pages/383123483/User+search+criteria

Note that this is only the API interface and that the functionality will still be implemented in a separate task.


AltusBarry:
+1

PR #76, opened 2018-04-26
Feature/ge 1017

AltusBarry:


authentication_service/integration.py
```diff
@@ -86,15 +88,66 @@ def user_list(request, offset=None, limit=None, birth_date=None, country=None, d
...
-        users = CoreUser.objects.values(*USER_VALUES).order_by("id")
-
-        if user_ids:
-            users = users.filter(id__in=user_ids)
+        order_by = order_by or ["id"]
+        users = get_user_model().objects.values(
+            *USER_VALUES).order_by(*order_by)
+
+        # Bools
+        if tfa_enabled:
```
cobusc: Check both conditions please. Something like:
```
if tfa_enabled is not None:
    if tfa_enabled is True:  # Just check that a bool is returned and not a string "true"
        users = users.filter(totpdevice__isnull=False)
    else:  # The case for False
        users = users.filter(totpdevice__isnull=True)
```

authentication_service/integration.py
```diff
@@ -86,15 +88,66 @@ def user_list(request, offset=None, limit=None, birth_date=None, country=None, d
...
+            users = users.filter(totpdevice__isnull=False)
+        if has_organisational_unit:
```
cobusc: See comment for `tfa_enabled`, please.

authentication_service/integration.py
```diff
@@ -86,15 +88,66 @@ def user_list(request, offset=None, limit=None, birth_date=None, country=None, d
...
+            users = users.filter(
+                organisational_unit__isnull=False
+                    if has_organisational_unit else True
+            )
+        if email_verified:
```
cobusc: This is not right. If `email_verified` is False, no filter will be applied, which is not what we want. Try this:
```
if email_verified is not None:
    users = users.filter(email_verified=email_verified)       # Assuming email_verified is a bool
```

authentication_service/integration.py
```diff
@@ -86,15 +88,66 @@ def user_list(request, offset=None, limit=None, birth_date=None, country=None, d
...
+            users = users.filter(email_verified=email_verified)
+        if is_active:
```
cobusc: See comment for `email_verified`

authentication_service/utils.py
```diff
@@ -112,3 +114,32 @@ def to_dict_with_custom_fields(instance, custom_fields):
             else:
                 result[field.name] = getattr(instance, field.name)
     return result
+
+def range_filter_parser(date_range):
+    parsed_range = []
+    if not isinstance(date_range, list):
+        # Depending on the format of the dates inside, literal_eval will break.
+        date_range = date_range.replace(" ", "").replace(
```
cobusc: Because this is a string, no validation is performed at the Swagger level. I'd like us to do proper parsing here (even if doing a regex match) to extract the information and raise an error if the format is incorrect, rather than break.

authentication_service/utils.py
```diff
@@ -112,3 +114,32 @@ def to_dict_with_custom_fields(instance, custom_fields):
...
+            "[", "").replace("]", "").split(",")
+
+    for date in date_range:
+        if date != "None":
+            # Assumptions are made about the date format, based on swagger spec.
+            # 2018-04-26T10:44:47.021Z
+            parsed_range.append(
+                datetime.datetime.strptime(date.split("T")[0], "%Y-%M-%d"))
```
cobusc: We need to support date and datetime fields, IIRC.

authentication_service/utils.py
```diff
@@ -112,3 +114,32 @@ def to_dict_with_custom_fields(instance, custom_fields):
...
+        else:
+            parsed_range.append(None)
+
+    # On the off chance there are more than 2 entries in the list.
+    if len(parsed_range) > 2:
```
cobusc: When using proper parsing this check is not necessary.

authentication_service/lookups.py
```diff
@@ -0,0 +1,14 @@
...
+
+@Field.register_lookup
+class Ilike(IContains):
+    lookup_name = "ilike"
+
+    def as_sql(self, compiler, connection):
+        lhs, lhs_params = self.process_lhs(compiler, connection)
+        rhs, rhs_params = self.process_rhs(compiler, connection)
+        params = lhs_params + rhs_params
+        return f"{lhs} ILIKE {rhs}", params
```
cobusc: Can we add the `%`s here? Like this: `return f"{lhs} ILIKE %{rhs}%", params`
That means it is never explicitly communicated over the API or handled our code. It is implied.

authentication_service/tests/test_integration.py
```diff
@@ -249,3 +267,127 @@ def test_user_delete(self):
...
+        self.assertEqual(len(response.json()), len(users))
+        response = self.client.get(
+            "/api/v1/users?birth_date="
+            "[2006-01-01T10:44:47.021Z,None]"
+        )
+        self.assertEqual(len(response.json()), len(users))
+        self.assertEqual(len(response.json()), len(users))
+        response = self.client.get(
+            "/api/v1/users?birth_date="
+            "[None, 1980-01-01T10:44:47.021Z]"
```
cobusc: Can we add a test for `[None,None]` too, please. It should return an HTTP 400 (Bad Request)

authentication_service/utils.py
```diff
@@ -112,3 +114,32 @@ def to_dict_with_custom_fields(instance, custom_fields):
...
+
+def range_filter_parser(date_range):
+    parsed_range = []
+    if not isinstance(date_range, list):
+        # Depending on the format of the dates inside, literal_eval will break.
+        date_range = date_range.replace(" ", "").replace(
+            "[", "").replace("]", "").split(",")
+
+    for date in date_range:
+        if date != "None":
```
cobusc: `if date.lower() != "none"`

authentication_service/integration.py
```diff
@@ -86,15 +88,66 @@ def user_list(request, offset=None, limit=None, birth_date=None, country=None, d
...
+            users = users.filter(q__ilike=q)
+
+        # Other filters
+        if country:
+            users = users.filter(country__code=country)
+        if user_ids:
+            users = users.filter(id__in=user_ids)
+        if gender:
+            users = users.filter(gender=gender)
+        if msisdn:
```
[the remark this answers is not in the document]
AltusBarry: msisdn has no index on it of any kind.

authentication_service/lookups.py
AltusBarry: This lookup subclasses IContain, the sql does include the %% eg:`authentication_service_coreuser"."username" ILIKE %SerNAme%`

authentication_service/integration.py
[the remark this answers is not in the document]
cobusc: It should: https://praekelt.atlassian.net/wiki/spaces/GEM/pages/383123483/User+search+criteria

authentication_service/lookups.py
cobusc: Kewl, that makes sense. Thanks.

authentication_service/tests/test_integration.py
```diff
@@ -249,3 +267,129 @@ def test_user_delete(self):
...
+        response = self.client.get(
+            "/api/v1/users?birth_date="
+            "[2007-01-01T10:44:47.021Z,2018-04-26T10:44:47.021Z]"
+        )
+        self.assertEqual(len(response.json()), len(users))
+        response = self.client.get(
+            "/api/v1/users?birth_date="
+            "[2006-01-01T10:44:47.021Z,None]"
+        )
+        self.assertEqual(len(response.json()), len(users))
```
cobusc: Double?

PR #77, opened 2018-05-01
Rework required for rebase on .org django bootstrap Docker image

cobusc:
Achtung! Currently the .org Django bootstrap image changes the ownership of the source directory and all files and directories underneath it when running in docker compose. This is definitely not ideal. I need to figure out if it is the celery instance or the auth service container doing this (for gunicorn and nginx), or whether it is both.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,22 +1,25 @@
-FROM praekeltfoundation/python-base:3.6
+FROM praekeltfoundation/django-bootstrap:py3
 
 ARG EXTRA_DEPS
 
 ENV DJANGO_SETTINGS_MODULE=project.settings
+ENV SKIP_MIGRATIONS=1
 
 # Git is required because one of the pip requirements is pulled from github.
 RUN apt-get update && apt-get install -y git gcc netcat $EXTRA_DEPS gettext libgettextpo-dev
 
-RUN mkdir /static
-
-WORKDIR /app/
-
-COPY ./requirements /app/requirements
-
-RUN pip3 install --no-cache-dir -r /app/requirements/requirements.txt --src /usr/local/src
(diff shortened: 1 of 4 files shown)


Dockerfile
```diff
@@ -1,22 +1,25 @@
-FROM praekeltfoundation/python-base:3.6
+FROM praekeltfoundation/django-bootstrap:py3
 
 ARG EXTRA_DEPS
 
 ENV DJANGO_SETTINGS_MODULE=project.settings
+ENV SKIP_MIGRATIONS=1
```
AltusBarry: Is 0 then to run them? Just checking for QA setup.

Dockerfile
```diff
@@ -1,22 +1,25 @@
...
 
 # Git is required because one of the pip requirements is pulled from github.
 RUN apt-get update && apt-get install -y git gcc netcat $EXTRA_DEPS gettext libgettextpo-dev
 
-RUN mkdir /static
+RUN mkdir -p /app/static
 
-WORKDIR /app/
+COPY . /app/
```
AltusBarry: Don't copy the app before we install requirements, if possible. Requirement layer will be cached unless the requirements change. But the actual app is likely to change each deploy and will bust all cached layers beneath it.

Dockerfile
cobusc: No. You need to omit the environment variable entirely. I'll remove this from here and add to the `docker-compose.yml` file in `core-general`.

Dockerfile
cobusc: The requirements are not installed under the `/app` directory, so it should not matter. It is installed in the global python scope and `/usr/local/src`.

PR #78, opened 2018-05-03
feature/GE-1017: Added range extras and tests

AltusBarry:


authentication_service/utils.py
```diff
@@ -115,31 +117,56 @@ def to_dict_with_custom_fields(instance, custom_fields):
...
+        if isinstance(date, str) and date.lower() != "none":
             # Assumptions are made about the date format, based on swagger spec.
             # 2018-04-26T10:44:47.021Z
-            parsed_range.append(
-                datetime.datetime.strptime(date.split("T")[0], "%Y-%M-%d"))
+            try:
+                parsed_range.append(
+                    datetime.datetime.strptime(
+                        date.split("T")[0], "%Y-%M-%d"
+                    ).date()
```
cobusc: Rather convert to datetime, if possible. DB should interpret it correctly.

fritzbrand:
+1 (Wahaha, as if that means anything)

PR #81, opened 2018-05-04
Disabled site login prevention

cobusc:


authentication_service/middleware.py
```diff
@@ -125,6 +125,18 @@ def process_view(self, request, view_func, view_args, view_kwargs):
                     pass
                 else:
                     raise e
+
+            if not api_helpers.is_site_active(authorize.client):
```
AltusBarry: Since this does an api call on, potentially, every GET. Should this not rather be for request url, `/login/`?

AltusBarry:
+1

PR #83, opened 2018-05-07
Changed date parser to handle a json type object as an argument with …

DavidOberholzer:
*Background*: The GMP date range filter can easily send off an object with either/or/both `to` and `from` attributes in the query parameters. The list route on AOR/ReduxForm is far more unpleasant and hacky.

*What was done*: Changed the date parser to handle a given object parsing it from JSON. Tests also updated.

### DIFF
--- a/authentication_service/utils.py
+++ b/authentication_service/utils.py
@@ -124,7 +124,7 @@ def range_filter_parser(date_range):
     parsed_range = []
     if not isinstance(date_range, list):
         # Depending on the format of the dates inside, literal_eval will break.
-        # Handle a string argument given if it looks like a string or a list.
+        # Handle a string argument given if it looks like an object or a list.
         if "{" in date_range:
             dates = date_range.replace('"', "").replace(
                 " ", "").replace("{", "").replace("}", "").split(",")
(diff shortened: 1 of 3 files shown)


authentication_service/utils.py
```diff
@@ -121,13 +120,22 @@ def to_dict_with_custom_fields(instance, custom_fields):
...
 
 def range_filter_parser(date_range):
+
     parsed_range = []
     if not isinstance(date_range, list):
         # Depending on the format of the dates inside, literal_eval will break.
-        date_range = date_range.replace(" ", "").replace(
-            "[", "").replace("]", "").split(",")
+        # Handle a string argument given if it looks like an object or a list.
+        if "{" in date_range:
```
cobusc: I'd prefer it if we parse it as JSON. The following is pseudo-code:
```
try:
    date_range = json.loads(date_range, encoding="utf-8")
except json.ValidationError as e:
   raise SuspiciousOperation(e.message)

from = date_range.get("from")  
if from:
    # Convert from to a date or time object here
to = date_range.get("to")
if to:
    # Convert to to a date or time object here

...
```

cobusc:
:+1: Thanks for this @DavidOberholzer

PR #85, opened 2018-05-15
GE-1036, et al.

cobusc:
The following was done:
* Added new UserSite model to keep track of which Sites a user has logged into via the Auth Service.
* Expose a new `site_ids` filter in the user listing endpoint, which forms the basis of site-specific and domain-specific user searches.
* Expose Country and OrganisationalUnit via the API for lookup purposes only. No create/delete/update functionality: GE-1051
* Fixed a bug: GE-1062


authentication_service/models.py
```diff
@@ -144,6 +146,21 @@ class Meta:
...
+    user = models.ForeignKey("CoreUser")
+    site_id = models.IntegerField()
+    consented_at = models.DateTimeField(auto_now_add=True)
+    created_at = models.DateTimeField(auto_now_add=True)
+    updated_at = models.DateTimeField(auto_now=True)
+
+    class Meta:
+        unique_together = ["user", "site_id"]
+        indexes = [
+            models.Index(fields=["user"]),
```
AltusBarry: Do you need an explicit index on foreignkeys? I think Django might default to one on them.

AltusBarry:
+1 only one question

authentication_service/models.py
cobusc: Well spotted. I should not need to define this explicitly. Thanks!

PR #92, opened 2018-05-18
Feature/ge 630

AltusBarry:
Largest part of migration flow.

Still requires app_id and site_id to be grabbed off of next query string. Also still requires user lookup to make use of them.


authentication_service/user_migration/views.py
```diff
@@ -0,0 +1,133 @@
...
+    ("userdata", UserDataForm),
+    ("securityquestions", forms.SecurityQuestionFormSet),
+)
+class MigrateUserWizard(views.LanguageMixin, NamedUrlSessionWizardView):
+    form_list = migration_forms
+    instance_dict = {
+        "securityquestions": models.SecurityQuestion.objects.none()
+    }
+
+    @generic_deprecation(
```
cobusc: Nice!

authentication_service/models.py
```diff
@@ -93,6 +94,7 @@ class CoreUser(AbstractUser):
         ],
         null=True
     )
+    migration_data = JSONField(blank=True, null=True)
```
cobusc: I usually prefer not allowing NULLs for JSONB columns, but rather explicitly defaulting to `{}` (an empty object).

authentication_service/user_migration/forms.py
```diff
@@ -0,0 +1,55 @@
...
+    password2 = forms.CharField(
+        label=_("Password confirmation"),
+        widget=forms.PasswordInput,
+        strip=False,
+        help_text=_("Enter the same password as before, for verification."),
+    )
+
+    def clean_username(self):
+        username = self.cleaned_data.get("username")
+        if get_user_model().objects.filter(username=username).count() > 0:
```
cobusc: You can use `exists()` instead of `count() > 0`.

authentication_service/user_migration/views.py
```diff
@@ -0,0 +1,133 @@
...
+                self.token,
+                salt="ge-migration-user-registration",
+                max_age=15*360 # 15 min in seconds
+            )
+        except signing.SignatureExpired:
+            messages.error(
+                self.request,
+                _("Migration url has expired, please login again.")
+            )
+            return redirect(self.get_login_url(query=query))
```
cobusc: Nice

authentication_service/user_migration/views.py
```diff
@@ -0,0 +1,133 @@
...
+
+        # Super sets up storage, needs to happen after token has been checked
+        dispatch = super(MigrateUserWizard, self).dispatch(*args, **kwargs)
+
+        # Wizard querystrings get stripped, this will ovalidate true once.
```
cobusc: Typo

authentication_service/user_migration/views.py
```diff
@@ -0,0 +1,133 @@
...
+        # Change get_step_url to persist querystrings.
```
cobusc: This comment does not seem to be true.

authentication_service/user_migration/forms.py
AltusBarry: :woman_facepalming: yeah, I forgot. Thanks.

authentication_service/user_migration/views.py
AltusBarry: It is, the wizard still strips querystrings by default. I will just make it clearer. We persist the querystring in the storage ourselves.

If we want the querystring to persist on the urls at all times, we will need to do this in get_step_url.

authentication_service/user_migration/views.py
```diff
@@ -0,0 +1,133 @@
...
+    )
+    def dispatch(self, *args, **kwargs):
+        self.token = self.kwargs["token"]
+
+        # Grab the query early, for use in the event the token has expired
+        query = self.request.GET.get("persist_query", None)
+
+        # Check if token has expired
+        try:
+            self.temp_id = signing.loads(
```
cobusc: Can we rather call this the `migration_user_id`?

authentication_service/user_migration/views.py
```diff
@@ -0,0 +1,133 @@
...
+                id=self.temp_id
+            )
+        except TemporaryMigrationUserStore.DoesNotExist:
+            raise Http404(
+                f"Migrating user with id {self.temp_id} does not exist."
+            )
+
+    def get_login_url(self, query=None):
+        query = self.storage.extra_data.get("persist_query", query)
+        next_query = f"?next={query}" if query is not None else ""
```
cobusc: Should we not rather omit the `next=` argument if the query is None?

authentication_service/views.py
```diff
@@ -11,8 +14,12 @@
 )
 from django.core.exceptions import ValidationError
 from django.core.urlresolvers import reverse, reverse_lazy
+
+from django.contrib.auth import get_user_model
+from django.core import signing
+# NOTE: Can be refactored, both redirect import perform more or less the same.
```
cobusc: Is this still applicable?

authentication_service/views.py
```diff
@@ -26,8 +33,13 @@
 from two_factor.views import core
 
 from authentication_service import forms, models, tasks, constants
+from authentication_service.decorators import generic_deprecation
 from authentication_service.forms import LoginForm
 
+# TODO catch import error
```
cobusc: Remove comment

authentication_service/user_migration/views.py
AltusBarry: We do, its either `"?next=/url/` or `""`

authentication_service/views.py
AltusBarry: Yes, just don't want to refactor as part of this work. So will do it as its own PR.

cobusc:
:+1: I think most of my comments were cosmetic in nature, so apply as you see fit.

authentication_service/user_migration/views.py
cobusc: Oh yes. Now I see.

cobusc:
:+1: again :)

PR #97, opened 2018-05-21
Updates to load multi-site config

cobusc:


authentication_service/management/commands/demo_content.py
```diff
@@ -34,15 +35,16 @@ def handle(self, *args, **options):
...
                 "client_secret": "super_client_secret_1",
                 "response_type": "code",
                 "jwt_alg": "HS256",
                 "redirect_uris": [
                     os.environ.get("WAGTAIL_1_CALLBACK", 'http://example.com/'),
                     os.environ.get(
                         "WAGTAIL_1_CALLBACK",
-                        'http://example.com/') + "register-redirect/"
+                        'http://example.com/') + "register-redirect/",
+                    os.environ.get("WAGTAIL_1_LOGOUT_REDIRECT", 'http://example.com/')
```
AltusBarry: Why is the logout redirect added to the `redirect_uri`s?

AltusBarry:
+1, one question

authentication_service/management/commands/demo_content.py
cobusc: The callback URLs look like this: `http://wagtail-demo-1-site-1:8000/oidc/callback/`
The logout URLs look like this: `http://wagtail-demo-1-site-1:8000/`, which happens to be the URLs that we redirect back to.

cobusc: I realise now that that "register-redirect/" URL looks incorrect. It computes to `http://wagtail-demo-1-site-1:8000/oidc/callback/register-redirect/` and from the `urls.py` in the `core-integration-demo` it looks like it should be `http://wagtail-demo-1-site-1:8000/register-redirect/`. I'll fix it.

PR #102, opened 2018-05-28
feature/GE-1088: Removed RedirectMixin, updated util method name

AltusBarry:
Remove RedirectMixin, mod views to still redirect as expected.

Updated util method name, to be in line with session data getter.


cobusc:
:+1: Thanks

PR #103, opened 2018-05-28
Feature/ge 1089

AltusBarry:
Refactor of Middleware.

Update to session data helper utils
Site is active split out to own middleware
Theme middleware sets and clears values on whitelisted urls
Session middleware sets and clears values on whitelisted urls

### DIFF
--- a/authentication_service/constants.py
+++ b/authentication_service/constants.py
@@ -1,18 +1,18 @@
-SECURITY_QUESTION_COUNT = 2
-
-MIN_NON_HIGH_PASSWORD_LENGTH = 4
-
 # Dictionary key within which we should store all our extra session data. Makes
 # cleanup much easier as we only need to remove one key. utils.update and get
 # data methods already make use of this.
 EXTRA_SESSION_KEY = "ge_session_extra_data"
 
-SESSION_KEYS = {
-    "redirect_client_uri": "ge_oidc_client_uri",
-    "redirect_client_name": "ge_oidc_client_name",
-    "redirect_client_terms": "ge_oidc_client_terms",
-    "theme": "theme"
-}
+
+class SessionKeys:
+    CLIENT_URI = "ge_oidc_client_uri"
+    CLIENT_NAME = "ge_oidc_client_name"
+    CLIENT_TERMS = "ge_oidc_client_terms"
+    THEME = "theme"
+
+SECURITY_QUESTION_COUNT = 2
+
+MIN_NON_HIGH_PASSWORD_LENGTH = 4
 
 GE_TERMS_URL = "https://www.girleffect.org/terms-and-conditions/"
 
@@ -26,5 +26,5 @@ def get_theme_client_name(request):
     fallback_name = request.session.get(
         EXTRA_SESSION_KEY,
         {}
-    ).get(SESSION_KEYS["redirect_client_name"])
+    ).get(SessionKeys.CLIENT_NAME)
     return THEME_NAME_MAP.get(theme, fallback_name)
--- a/authentication_service/constants.py
+++ b/authentication_service/constants.py
@@ -22,7 +22,7 @@ THEME_NAME_MAP = {
 
 
 def get_theme_client_name(request):
-    theme = request.META["X-Django-Layer"]
+    theme = request.META.get("X-Django-Layer", None)
     fallback_name = request.session.get(
         EXTRA_SESSION_KEY,
         {}
--- a/authentication_service/middleware.py
+++ b/authentication_service/middleware.py
@@ -16,7 +16,9 @@ from django.utils.translation import ugettext as _
 
 from authentication_service import exceptions, api_helpers
 from authentication_service.constants import SESSION_KEYS, EXTRA_SESSION_KEY
-from authentication_service.utils import update_session_data, get_session_data
+from authentication_service.utils import (
+    update_session_data, get_session_data, delete_session_data
+)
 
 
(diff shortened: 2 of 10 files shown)


authentication_service/middleware.py
```diff
@@ -43,55 +77,47 @@ def fetch_theme(request, key=None):
             # list.
             theme = next_query_args.get("theme", [None])[0]
 
-    if theme is None:
-        theme = get_session_data(request, key)
-
-    return theme
+    return theme.lower() if isinstance(theme, str) else theme
```
cobusc: What can `theme` be other than a string?

authentication_service/middleware.py
```diff
@@ -16,11 +16,45 @@
...
 from authentication_service.constants import SESSION_KEYS, EXTRA_SESSION_KEY
-from authentication_service.utils import update_session_data, get_session_data
+from authentication_service.utils import (
+    update_session_data, get_session_data, delete_session_data
+)
 
 
 LOGGER = logging.getLogger(__name__)
 
+SESSION_UPDATE_URL_WHITELIST = [
```
cobusc: Maybe use a `set` for constant time `in` lookups.

authentication_service/middleware.py
```diff
@@ -43,55 +77,47 @@ def fetch_theme(request, key=None):
...
-            except AuthorizeError as e:
-                # Suppress one of the errors oidc raises. It is not required
-                # for pages beyond login.
-                if e.error == "unsupported_response_type":
-                    pass
-                else:
-                    raise e
-
+            authorize = authorize_client(request)
+            if not isinstance(authorize, AuthorizeEndpoint):
```
cobusc: I prefer the positive check here: `if isinstance(authorize, HttpResponse):`

authentication_service/tests/settings/111.py
```diff
@@ -82,7 +82,6 @@
     "django.middleware.locale.LocaleMiddleware",
     "django_otp.middleware.OTPMiddleware",
     "authentication_service.middleware.ErrorMiddleware",
-    "authentication_service.middleware.ThemeManagementMiddleware",
```
cobusc: Why was this removed from tests?

project/settings.py
```diff
@@ -88,7 +88,6 @@
     "django.middleware.locale.LocaleMiddleware",
     "django_otp.middleware.OTPMiddleware",
     "authentication_service.middleware.ErrorMiddleware",
-    "authentication_service.middleware.ThemeManagementMiddleware",
```
cobusc: Why is this removed?

cobusc: Oh, I see it moved to base.

authentication_service/tests/settings/111.py
cobusc: Oh, I see it moved to base.

cobusc:
Left some comments

authentication_service/middleware.py
AltusBarry: Should not be anything other than a string. Will explicitly cast to string.

authentication_service/middleware.py
```diff
@@ -104,69 +130,73 @@ def create_disabled_site_response(request):
...
                 )
 
+
+class SessionDataManagementMiddleware(MiddlewareMixin):
+    """
+    NOTE: This Middleware should always be as near the end of the Middleware
+    list in settings. Middleware is evaluated in order and this needs to happen
+    as near the end as possible.
+    """
+    client_uri_key = SESSION_KEYS["redirect_client_uri"]
```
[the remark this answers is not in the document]
AltusBarry: I like that. Thanks.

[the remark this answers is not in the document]
AltusBarry: These keys aren't specifically for this middleware class or middleware in general. They are just keys that need to stay constant, hence `constants.py`.

The values can potentially be set else where too. This middleware just happens to be the only place that sets them.

Making use of a constant dictionary is a habit from having had to make use of shared keys between many different Django apps. If you'd prefer an object vs a dictionary, that can be arranged. However I will keep it in `constants.py`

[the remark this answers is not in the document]
cobusc: Yes, the crux was that it is not necessary to look up the name of something you want using a hardcoded string value. Then you might as well hardcode the looked up value. So it is the use of a dictionary that is not nice IMHO.
I'm very happy that these are defined in `constants.py`

authentication_service/constants.py
```diff
@@ -1,18 +1,18 @@
...
 EXTRA_SESSION_KEY = "ge_session_extra_data"
 
-SESSION_KEYS = {
-    "redirect_client_uri": "ge_oidc_client_uri",
-    "redirect_client_name": "ge_oidc_client_name",
-    "redirect_client_terms": "ge_oidc_client_terms",
-    "theme": "theme"
-}
+
+class SessionKeys:
```
cobusc: Thanks

authentication_service/views.py
```diff
@@ -40,7 +40,7 @@
 from authentication_service.user_migration.models import TemporaryMigrationUserStore
 
 
-REDIRECT_SESSION_KEY = constants.SESSION_KEYS["redirect_client_uri"]
+REDIRECT_SESSION_KEY = constants.SessionKeys.CLIENT_URI
```
cobusc: Please change this variable name to be something more appropriate (e.g. `CLIENt_URI`)

cobusc:
:+1: when last nitpick is sorted out.

PR #104, opened 2018-05-29
Feature/ge 1073

AltusBarry:
Add client_id lookup for migration of users.


authentication_service/oidc_provider_settings.py
```diff
@@ -98,9 +98,13 @@ def scope_site(self) -> dict:
         data = api_helpers.get_user_site_data(
             self.user.id, site_id).to_dict()["data"]
         now = timezone.now().astimezone(datetime.timezone.utc).isoformat()
+        # TODO Only send migration_information along if client_ids match
```
cobusc: This comment can be removed.

authentication_service/oidc_provider_settings.py
```diff
@@ -98,9 +98,13 @@ def scope_site(self) -> dict:
...
         result = {
             "site": {"retrieved_at": f"{now}", "data": data},
-            "migration_information": self.user.migration_data
+            "migration_information": self.user.migration_data \
```
cobusc: Alternative:
```python
    result = {
￼        "site": {"retrieved_at": f"{now}", "data": data}
    }
    if self.client.client_id == self.user.migration_data.get("client_id"):
        result["migration_information"] = self.user.migration_data
```

authentication_service/user_migration/models.py
```diff
@@ -9,22 +9,36 @@ class TemporaryMigrationUserStore(models.Model):
         help_text="Required. 150 characters or fewer. Letters, digits and @/./+/-/_ only.",
     )
     pw_hash = models.CharField("password", max_length=128)
-    app_id = models.IntegerField()
     user_id = models.IntegerField()
-    site_id = models.IntegerField()
+    client_id = models.CharField(max_length=255)
```
cobusc: Would it make sense to define this as foreign key? That way we avoid loading bad data.

cobusc:
:+1: One question and one suggestion.

cobusc:
:+1: :1st_place_medal:

PR #109, opened 2018-06-04
Feature/ge 1098

AltusBarry:
Added new locales

Added middleware subclassing Django LocalMiddleware, that will attempt to redirect to the language requested via querystring or 404 as intended.


authentication_service/middleware.py
```diff
@@ -224,3 +231,59 @@ class ErrorMiddleware(MiddlewareMixin):
...
+
+    def process_request(self, request):
+        super(GELocaleMiddleware, self).process_request(request)
+
+        # Get the language code to use
+        language_code = request.GET.get(
+            LANGUAGE_QUERY_PARAMETER, None
+        )
+
+
```
cobusc: Unnecessary whitespace

authentication_service/middleware.py
```diff
@@ -224,3 +231,59 @@ class ErrorMiddleware(MiddlewareMixin):
...
+    Overrides the default logic to ALWAYS attempt to switch to the language
+    provided via querystring.
+
+    Most of this code was inspired by
+    django.middleware.locales.LocaleMiddleware itself.
+    """
+
+    def process_request(self, request):
+        super(GELocaleMiddleware, self).process_request(request)
```
cobusc: Why does it feel like this should either be at the end of this function, i.e. only called if no redirect is done?

cobusc:
:+1: Nice work!

authentication_service/middleware.py
AltusBarry: The super also activates a language. So if we are already on the correct language we need not redirect.

PR #110, opened 2018-06-04
Added healthcheck API end-point

cobusc:
* Also pinned Django version 1.11.12. For some reason 1.11.3 breaks custom primary key "code" on country table.

### DIFF
--- a/authentication_service/views.py
+++ b/authentication_service/views.py
@@ -1,4 +1,8 @@
+import datetime
+import socket
 from datetime import date
+
+import pkg_resources
 from dateutil.relativedelta import relativedelta
 from urllib.parse import urlparse, parse_qs
 import urllib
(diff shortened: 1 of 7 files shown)


authentication_service/views.py
```diff
@@ -498,3 +503,21 @@ def get_success_url(self):
 
 class PasswordResetConfirmView(PasswordResetConfirmView):
     form_class = forms.SetPasswordForm
+
+
+class HealthCheckView(View):
```
AltusBarry: Set allowed methods on this view(`http_method_names=["get"]`), will stop anything other than GET being allowed.

AltusBarry:
+1, just the one comment. Its more of a suggestion than anything else.

AltusBarry:
+1, ignore comment. Base Django view already raises a method not allowed if the method was not added to the view.

authentication_service/views.py
cobusc: As per your own comment below (copied here for reference): "Base Django view already raises a method not allowed if the method was not added to the view."

PR #111, opened 2018-06-05
Unpin Django version from 1.11.12

cobusc:
...so that we can use the latest bugfix release (which is 1.11.13 at the time of writing).

We had an issue that is resolved when using a squashed migration, since it prunes some definitions that were add and then later removed.


AltusBarry:
+1, remember to then delete the old migrations and make the required changes and redeploy.

PR #113, opened 2018-06-05
Feature/ge 1085

AltusBarry:


authentication_service/constants.py
```diff
@@ -8,6 +8,7 @@ class SessionKeys:
     CLIENT_URI = "ge_oidc_client_uri"
     CLIENT_NAME = "ge_oidc_client_name"
     CLIENT_TERMS = "ge_oidc_client_terms"
+    CLIENT_ID = "ge_client_django_id"
```
cobusc: Why `django_id`? "ge_oidc_client_id" would be clearer and follow the convention of the other keys.

authentication_service/templates/authentication_service/form.html
```diff
@@ -0,0 +1,27 @@
+{% extends "base.html" %}
```
cobusc: It is not clear from the name of the file what form this is, only that it is a form.

authentication_service/templates/ninyampinga/authentication_service/form.html
```diff
@@ -0,0 +1,8 @@
+{% extends "authentication_service/form.html" %}
```
cobusc: It is not clear from the name of the file what form this is, only that it is a form.

authentication_service/templates/springster/authentication_service/form.html
```diff
@@ -0,0 +1,8 @@
+{% extends "authentication_service/form.html" %}
```
cobusc: Ditto.

authentication_service/templates/zathu/authentication_service/form.html
```diff
@@ -0,0 +1,8 @@
+{% extends "authentication_service/form.html" %}
```
cobusc: Ditto

authentication_service/user_migration/views.py
```diff
@@ -132,3 +136,119 @@ def get_user_data(self):
     def get_login_url(self, query=None):
         query = self.storage.extra_data.get("persist_query", query)
         return redirect(query or reverse("login"))
+
+
+class QuestionGateView(FormView):
```
cobusc: We should protect this view using Django defender.

authentication_service/user_migration/models.py
```diff
@@ -31,17 +31,19 @@ class Meta:
     def check_password(self, raw_password):
         return check_password(raw_password, self.pw_hash)
 
-    def check_answers(self, answer_one, answer_two):
-        return all([
-            check_password(answer_one, self.answer_one),
-            check_password(answer_two, self.answer_two)
-        ])
+    def check_answer_one(self, answer):
+        return check_password(answer.lower(), self.answer_one)
```
cobusc: `answer.strip().lower()` maybe?

authentication_service/user_migration/models.py
```diff
@@ -31,17 +31,19 @@ class Meta:
...
+
+    def check_answer_two(self, answer):
+        return check_password(answer.lower(), self.answer_two)
```
cobusc: `answer.strip().lower()` maybe?

authentication_service/user_migration/models.py
```diff
@@ -31,17 +31,19 @@ class Meta:
...
     def set_password(self, raw_password):
         self.pw_hash = make_password(raw_password)
         self.save()
 
-    def set_anwers(self, raw_password):
-        self.answer_one = make_password(answer_one)
-        self.answer_two = make_password(answer_two)
+    def set_anwers(self, answer_one=None, answer_two=None):
+        if answer_one:
+            self.answer_one = make_password(answer_one.lower())
```
cobusc: `answer_one.strip().lower()` maybe?

authentication_service/user_migration/models.py
```diff
@@ -31,17 +31,19 @@ class Meta:
...
+        if answer_two:
+            self.answer_two = make_password(answer_two.lower())
```
cobusc: `answer_two.strip().lower()` maybe?

authentication_service/user_migration/views.py
```diff
@@ -132,3 +136,119 @@ def get_user_data(self):
...
+        try:
+            self.migration_user_id = signing.loads(
+                self.token,
+                salt="ge-migration-user-pwd-reset",
+                max_age=10*360 # 10 min in seconds
+            )
+        except signing.SignatureExpired:
+            messages.error(
+                self.request,
+                _("Password reset url has expired, please login again.")
```
cobusc: Maybe change `please login again` to something like `please restart the password reset process`.

authentication_service/user_migration/views.py
```diff
@@ -132,3 +136,119 @@ def get_user_data(self):
...
+            )
+            return redirect(self.get_login_url())
+        return super(QuestionGateView, self).dispatch(*args, **kwargs)
+
+    def get_form_kwargs(self):
+        kwargs = super(QuestionGateView, self).get_form_kwargs()
+        language = translation.get_language()
```
cobusc: These 4 lines can be reduced to 2. No need for intermediate variables.

authentication_service/user_migration/views.py
```diff
@@ -132,3 +136,119 @@ def get_user_data(self):
...
+        user = self.get_user_data
+        kwargs["user"] = user
+        kwargs["language"] = language
+        return kwargs
+
+    def form_valid(self, form):
+        return self.get_success_url()
+
+    @cached_property
+    def get_user_data(self):
```
cobusc: Because this is a property, the `get_` prefix is unnecessary. `migration_user_data` would be nice.

authentication_service/user_migration/views.py
```diff
@@ -132,3 +136,119 @@ def get_user_data(self):
...
+        user = self.get_user_data
+        kwargs["user"] = user
+        return kwargs
+
+    def form_valid(self, form):
+        form.update_password()
+        return self.get_success_url()
+
+    @cached_property
+    def get_user_data(self):
```
cobusc: Because this is a property, the `get_` prefix is unnecessary. `migration_user_data` would be nice.

authentication_service/constants.py
AltusBarry: Because it's the django client.id, not client.client_id. Thought being more explicit will reduce confusion in the future.

authentication_service/templates/authentication_service/form.html
AltusBarry: That is exactly what it is, its just a template with a form, can be reused whenever we only have a form view and need nothing custom added.

authentication_service/user_migration/views.py
AltusBarry: Makes sense.

authentication_service/user_migration/views.py
AltusBarry: I know, it is a mistake that has been made previously on other views. I kept it consistent with the old usages of this property.

I'll update them all accordingly, rather than continue to make use of `get_`

AltusBarry: Seems there are no others. I think I'm loosing it.

authentication_service/constants.py
```diff
@@ -8,6 +8,7 @@ class SessionKeys:
     CLIENT_URI = "ge_oidc_client_uri"
     CLIENT_NAME = "ge_oidc_client_name"
     CLIENT_TERMS = "ge_oidc_client_terms"
+    CLIENT_ID = "ge_oidc_client_django_id"
```
cobusc: This is actually referred to as the `client_token_id` elsewhere in the code (API). Can we use `ge_oidc_client_token_id` here please?

authentication_service/templates/authentication_service/form.html
cobusc: Ah, a generic form template. Got it.

authentication_service/user_migration/views.py
cobusc: Did you forget to change this?

AltusBarry: No, I think because I split the line something went wrong with the change detection. Defos updated it.
https://github.com/girleffect/core-authentication-service/pull/113/commits/edfbe64d642995f374922753bdc1e4120539080c

authentication_service/constants.py
AltusBarry: Let me rather go double check the exact uses and change them all to whatever makes more sense.

PR #114, opened 2018-06-05
Added country loading management command

DavidOberholzer:
Added management command to load all countries into auth service.


authentication_service/management/commands/load_countries.py
```diff
@@ -0,0 +1,270 @@
...
+    help = "Create all SiteDataSchemas for all sites found on Access Control."
+
+    def handle(self, *args, **options):
+        for code, name in ISO3166.items():
+            self.stdout.write(
+                self.style.SUCCESS("Adding country %s" % name))
+            country, created = Country.objects.update_or_create(
+                code=code, defaults={"name": name}
+            )
+            if created:
```
cobusc: `if not created:`

cobusc:
One minor bugfix, then it is a :+1: Thanks for this @DavidOberholzer

PR #118, opened 2018-06-13
feature/migration-app-fixes: Fix past transgressions

AltusBarry:


authentication_service/user_migration/views.py
```diff
@@ -209,7 +211,7 @@ def get_success_url(self, query=None):
         return redirect(url)
 
     def get_login_url(self, query=None):
```
cobusc: `query` argument is never used.

authentication_service/user_migration/views.py
```diff
@@ -234,12 +236,13 @@ def dispatch(self, *args, **kwargs):
...
                 salt="ge-migration-user-pwd-gate-passed",
-                max_age=5*360 # 5 min in seconds
+                max_age=5*360  # 5 min in seconds
             )
         except signing.SignatureExpired:
             messages.error(
                 self.request,
-                _("Password reset url has expired, please login again.")
+                _("Password reset url has expired,"
+                    " please restart the password reset proces.")
```
cobusc: Odd indent

authentication_service/user_migration/views.py
```diff
@@ -267,3 +270,6 @@ def migration_user(self):
 
     def get_success_url(self, query=None):
         return redirect(reverse("password_reset_done"))
+
+    def get_login_url(self, query=None):
```
cobusc: `query` argument never used.

cobusc:
:+1: Just check if some of those unused arguments cannot be removed.

authentication_service/user_migration/views.py
AltusBarry: Added them in, preparation for an issue I know we will be dealing with. This flow does not drop the user back on login with the next parameter still present. The entire pwd reset flow doesn't cater for that, not just migration user pwd reset.

authentication_service/user_migration/views.py
AltusBarry: I blindly let flake8 tell me what to do today! Glad it looks odd to someone else too. Will revert rather.

PR #119, opened 2018-06-13
Added a block to override the custom class of the page

markupguy:
@AltusBarry pls

### DIFF
--- a/authentication_service/templates/base.html
+++ b/authentication_service/templates/base.html
@@ -39,7 +39,7 @@
     </head>
 
     <body>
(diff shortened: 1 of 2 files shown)


authentication_service/templates/base.html
```diff
@@ -39,7 +39,7 @@
     </head>
 
     <body>
-        <div id="wrap" class="{{ ge_global_theme|title }}{{ request.path_info|classify:"/" }}">
+        <div id="wrap" class="{% block custom-class-override %}{{ ge_global_theme|title }}{{ request.path_info|classify:"/" }}{% endblock %}">
```
AltusBarry: Should it override the entire class or rather enhance it?

AltusBarry:
+1

PR #120, opened 2018-06-13
feature/GE-1118: 404 if question language is not available

AltusBarry:


authentication_service/user_migration/views.py
```diff
@@ -169,6 +170,32 @@ def dispatch(self, *args, **kwargs):
...
+                    "No question translation matching the"
+                    " current language could be found."
+                )
+            },
+            status=404
+        )
+
+        # There is no guarantee both questions are present, check both
+        # separately.
+        if user.question_one:
```
cobusc:
```python
if user.question_one and not user.question_one.get(language, None) \
or user.question_two and not user.question_two.get(language, None):
    # Only construct the 404 response if necessary
    return TemplateResponse(...)
```

cobusc:
:+1: Left a suggestion

PR #123, opened 2018-06-14
Language mods

cobusc:
# What was done
1. Added convenience code in the `Makefile` for language locale generation.
2. Generated `.po` files for the languages we want to support at launch.
3. Modified `.gitignore` to ignore `.po` files under the tests directory, except for the `prs` language, which is used in a test.
4. Removed the following from `.po` files, which caused Django to break:
```
"Plural-Forms: nplurals=INTEGER; plural=EXPRESSION;\n"
```


AltusBarry:
+1, looks good.

PR #124, opened 2018-06-14
feature/registration-regression-fix: Registration was NOT working as …

AltusBarry:
…expected when no redirects were present

### DIFF
--- a/authentication_service/views.py
+++ b/authentication_service/views.py
@@ -242,9 +242,9 @@ class RegistrationView(LanguageMixin, CreateView):
 
         # If get_success_url already returns a response object, immediately
         # return it.
-        success_response = self.get_success_url()
-        if isinstance(success_response, HttpResponse):
-            return success_response
+        alternate_response = self.get_success_url()
+        if isinstance(alternate_response, HttpResponse):
+            return alternate_response
         return response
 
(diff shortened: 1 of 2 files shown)


authentication_service/views.py
```diff
@@ -239,6 +240,10 @@ def form_valid(self, form):
                 data["language_code"] = self.language
                 question = models.UserSecurityQuestion.objects.create(**data)
 
+        # If get_success_url already returns a response object, immediately
+        # return it.
+        if not isinstance(self.get_success_url(), str):
```
cobusc: I'd prefer if the function was only called once.

cobusc: Also, if the response is a `str` then what does that mean? Do we just ignore the value?

AltusBarry: The issue is, get_success_url should not return anything other than a str. Te more explicit check would be `if isinstance(self.get_success_url(), HttpResponse)`.

cobusc: I see. So it is the fact that `get_success_url` has been overloaded to do a `render()` as well.
Will we then ever have a use for `response`? Should we then have
```
result = self.get_success_url()
if isinstance(result, HttpResponse):
    return result
return redirect(result)
```

PR #137, opened 2018-06-26
Feature/ge 1086

AltusBarry:
Allow security questions on registration to be pre-selected via querystring parameters.

### DIFF
--- a/authentication_service/tests/test_views.py
+++ b/authentication_service/tests/test_views.py
@@ -977,6 +977,29 @@ class TestRegistrationView(TestCase):
             f'<option value="{self.question_three.id}" selected>{self.question_three.question_text}</option>'
         )
 
+    def test_question_preselect_incorrect_id(self):
+        # Test with redirect URI set.
+        response = self.client.get(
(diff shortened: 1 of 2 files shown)


authentication_service/tests/test_views.py
```diff
@@ -952,6 +961,45 @@ def test_view_terms_html(self):
...
+        self.assertContains(
+            response,
+            '<option value="" selected>---------</option>',
+            count=2
+        )
+        self.assertContains(
+            response,
+            "selected>",
+            count=3
+        )
```
DavidOberholzer: Small question for me, Why you checking this?

AltusBarry: Just to ensure nothing else was selected. If there was, there would be 4 selected elements.

There are 4 select fields on the page: Country, Gender, 2 questions. It's a peace of mind check.

One thing to note, the id is not one of the field values. If there was no number supplied the default "" value element WOULD actually be selected. With nothing selected browsers will still just send an empty string along and it will validate as usual.

PR #138, opened 2018-07-02
Fixed total count bug

cobusc:
# Background
https://praekelt.atlassian.net/browse/GE-1130

# What was done:
To avoid the count being incorrect, I had to remove the implied join with the UserSite table and replace it with a subquery. I tried looking at various other approaches and this is the only one that I could actually implement.


AltusBarry:
+1

PR #140, opened 2018-07-03
Feature/ge 1066

AltusBarry:
BLOCKED(Technically): Until feed back is provided on the first comment of https://praekeltorg.atlassian.net/browse/GEINFRA-64

Registration split into a two page wizard, skipping second step if user has an email address.


authentication_service/forms.py
```diff
@@ -68,9 +73,10 @@ class Meta:
         ]
         exclude = ["terms",]
 
-    def __init__(self, security=None, required=None, hidden=None, *args, **kwargs):
+    def __init__(self, term_url=None, security=None, required=None, hidden=None, *args, **kwargs):
```
cobusc: `terms_url` would be clearer and more consistent with `GE_TERMS_URL`.

authentication_service/tests/test_views.py
```diff
@@ -883,23 +920,31 @@ def test_redirect_view(self):
...
 
     def test_incorrect_required_field_logger(self):
         test_output = [
-            "WARNING:authentication_service.forms:"
-            "Received required field that is "
-            "not on form: someawesomefield",
-            "WARNING:authentication_service.forms:"
-            "Received required field that is "
-            "not on form: notontheform"
+            'WARNING:authentication_service.forms:Received required field that is not on form: notontheform',
```
cobusc: Why are these repeated 7 times?
It can be simplified as 
```
`["WARNING:authentication_service.forms:Received required field that is not on form: notontheform"] * 7 +
 ["WARNING:authentication_service.forms:Received required field that is not on form: someawesomefield"] * 7
```

authentication_service/views.py
```diff
@@ -162,82 +165,106 @@ def post(self, *args, **kwargs):
...
+)
+
+def show_security_questions(wizard):
+    cleaned_data = wizard.get_cleaned_data_for_step(
+        "userdata") or {"email": None}
+    return cleaned_data["email"] is None
+
+class RegistrationWizard(LanguageMixin, NamedUrlSessionWizardView):
+    form_list = registration_forms
+    file_storage = DefaultStorage()
```
cobusc: Why is file storage required?

authentication_service/tests/test_views.py
AltusBarry: Adding a condition_dict to a wizard causes forms to initialise more than once. We log these during initialisation.

authentication_service/views.py
```diff
@@ -162,82 +165,106 @@ def post(self, *args, **kwargs):
...
-                form=form, question_formset=formset)
-            )
-
-        # Let the user model save.
-        response = super(RegistrationView, self).form_valid(form)
-
-        # GE-1065 requires ALL users to be logged in
-        new_user = authenticate(username=form.cleaned_data["username"],
-                                password=form.cleaned_data["password1"])
+        ## GE-1065 requires ALL users to be logged in
```
cobusc: Why `##`?

authentication_service/views.py
```diff
@@ -162,82 +165,106 @@ def post(self, *args, **kwargs):
...
+        new_user = authenticate(username=username,
+                                password=pwd)
         login(self.request, new_user)
 
-        # Do some work and assign questions to the user.
+        ## Do some work and assign questions to the user.
```
cobusc: Why `##`?

authentication_service/views.py
AltusBarry: Probably block commented all the code and never un-commented the comment.

authentication_service/views.py
```diff
@@ -162,82 +165,106 @@ def post(self, *args, **kwargs):
...
-
-        # Either a new formset instance or an existing one is passed to the
-        # formset class.
-        if kwargs.get("question_formset"):
-            ct["question_formset"] = kwargs["question_formset"]
-        else:
-            ct["question_formset"] = self.get_formset
-        return ct
+    def done(self, form_list, **kwargs):
+        security_obj = types.SimpleNamespace()
```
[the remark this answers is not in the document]
AltusBarry: Late day brain. There's a better way to achieve what I wanted, will update.

authentication_service/views.py
cobusc: If we introduce filesystem storage we need to ensure that we have session affinity on the load balancers. I am not sure if this is currently the case.

cobusc: After chatting with the .org team it is clear that we cannot use something like session affinity. This means that any storage we use needs to be shared. This was suggested in case we want to use S3: https://github.com/jschneier/django-storages

Another option would be to store temporary data in Redis. Redis string (byte) values can be up to 512MB (https://redis.io/topics/data-types)

cobusc:
:+1:

PR #143, opened 2018-07-05
feature/GEINFRA-62: Remove birth_date from form html output

AltusBarry:
Edit profile was on face value not working. Turns out it was working for all fields apart from age. This was due to birth_date being passed an initial value and then being hidden on form the browser then ends up posting back.

There was a recent change to age and birth_date logic where all the forms were made to have the same basic functionality regarding those two fields. With that logic came change to edit profile where it no longer assumes that age is the only field present and birth_date value always takes priority when updating the user data.

### DIFF
--- a/authentication_service/forms.py
+++ b/authentication_service/forms.py
@@ -467,7 +467,8 @@ class EditProfileForm(forms.ModelForm):
         ]
 
     def _html_output(self, *args, **kwargs):
-        # Django does not allow the exclusion of fields on ModelForm forms.
+        # Exclude fields from the html not the form itself. Makes using built
+        # in save method easier.
 
         # Currently birth_date should not be merely hidden, it causes the
         # browser to send back the original value. Birth date is always used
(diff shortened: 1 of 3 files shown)


authentication_service/forms.py
```diff
@@ -467,6 +466,24 @@ class Meta:
             "age", "birth_date", "country", "avatar"
         ]
 
+    def _html_output(self, *args, **kwargs):
+        # Django does not allow the exclusion of fields on none ModelForm forms.
```
cobusc: This is a ModelForm-based form. I don't understand the comment,

authentication_service/forms.py
```diff
@@ -467,6 +466,24 @@ class Meta:
...
+
+        # Currently birth_date should not be merely hidden, it causes the
+        # browser to send back the original value. Birth date is always used
+        # for user age over the actual age field.
+        if self.fields["birth_date"].required is False \
```
cobusc: Nice solution.

cobusc:
:+1:

authentication_service/forms.py
AltusBarry: Thanks.

Will update, I changed the sentence a few times, seems I need to practise my proof reading skills again. Used to read "Allows the exclusion of fields only on none ModelForm forms.`

PR #146, opened 2018-07-06
Exclude language variants

cobusc:
# Background
We cannot support translations for all variations of English (American/British/Australian) and other language variants, so we limit languages to the base ones.

### DIFF
--- a/project/settings.py
+++ b/project/settings.py
@@ -358,7 +358,7 @@ if not any([IS_WORKER, env.bool("BUILDER", False)]):
         )
     )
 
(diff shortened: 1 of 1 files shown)


project/settings.py
```diff
@@ -341,7 +358,7 @@
         )
     )
 
-# SECURITY WARNING: don't run with debug turned on in production!
+# SECURITY WARNING: don"t run with debug turned on in production!
```
AltusBarry: Find replace failure

AltusBarry:
+1

PR #148, opened 2018-07-11
Send invitation email API end-point

cobusc:
See GEINFRA-66 for more info.

I also regenerated the client code for the Access Control component which was updated.


bootstrap.md
```diff
@@ -0,0 +1,28 @@
+# Create initial superuser
```
AltusBarry: Shouldn't this rather be in README.md or in the docs directory?

cobusc: True. I realised that I never committed it anywhere and just pulled it into the repo.
I think I'll link to it from the README. It is useful to know how I got the prod system bootstrapped, but it should not be necessary to repeat this, so it is for posterity.

PR #149, opened 2018-07-11
Feature/purge expired invitations

DavidOberholzer:
Added in endpoint to fire off purging expired invitations task on the access control. Celery workers now need access_control API settings, minor mocked test written as well.


authentication_service/api/urls.py
```diff
@@ -13,6 +13,7 @@
     url(r"^users$", views.Users.as_view()),
     url(r"^organisations/(?P<organisation_id>.+)$", views.OrganisationsOrganisationId.as_view()),
     url(r"^organisations$", views.Organisations.as_view()),
+    url(r"^ops/purge_expired_invitations_task$", views.OpsPurgeExpiredInvitationsTask.as_view()),
```
cobusc: @DavidOberholzer what do you think of making this `^invitations/purge_expired$` instead?

project/settings.py
```diff
@@ -317,17 +317,34 @@
...
     # GE API settings and setup
     ALLOWED_API_KEYS = env.list("ALLOWED_API_KEYS")
     USER_DATA_STORE_API_URL = env.str("USER_DATA_STORE_API")
     USER_DATA_STORE_API_KEY = env.str("USER_DATA_STORE_API_KEY")
-    ACCESS_CONTROL_API_URL = env.str("ACCESS_CONTROL_API")
-    ACCESS_CONTROL_API_KEY = env.str("ACCESS_CONTROL_API_KEY")
-
-    ## Setup API clients
+    # Setup API clients
     config = user_data_store.configuration.Configuration()
```
cobusc: We might as well move the `user_data_store` API stuff out of this `if` block too. Tht way we have more consistency between the configurations on the QA and prod environment components (celery workers and django daemons).
`ALLOWED_API_KEYS` only makes sense for the Django server, so it can stay.

swagger/authentication_service.yml
```diff
@@ -957,6 +964,19 @@ paths:
       tags:
         - authentication
 
+  /ops/purge_expired_invitations_task:
```
cobusc: As mentioned above, maybe change this to `/invitations/purge_expired`.

cobusc:
:+1:

PR #150, opened 2018-07-12
Added create/update/delete endpoints for organisations. Integration a…

DavidOberholzer:
Added create/update/delete endpoints for organisations. Integration ddded and tests.


AltusBarry:
+1 looks good

PR #151, opened 2018-07-12
Added check for users linked to Organisation, raise 400 if true. Adde…

DavidOberholzer:
Added check for users linked to Organisation, raise 400 if true. Added tests

### DIFF
--- a/authentication_service/integration.py
+++ b/authentication_service/integration.py
@@ -229,6 +229,11 @@ class Implementation(AbstractStubClass):
         organisation = get_object_or_404(Organisation, id=organisation_id)
         users = CoreUser.objects.filter(organisation=organisation)
         if users:
+            LOGGER.warning(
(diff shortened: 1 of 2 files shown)


authentication_service/integration.py
```diff
@@ -228,7 +227,11 @@ def organisation_delete(request, organisation_id, *args, **kwargs):
         :type organisation_id: integer
         """
         organisation = get_object_or_404(Organisation, id=organisation_id)
-        organisation.delete()
+        users = CoreUser.objects.filter(organisation=organisation)
+        if users:
+            raise BadRequestException
```
AltusBarry: Any chance we can add custom message to the exception? Might be useful to have user id and org id or something visible when we check logs.

AltusBarry:
+1, only made a suggestion

PR #152, opened 2018-07-13
Feature/geinfra 78

AltusBarry:


project/settings.py
```diff
@@ -379,3 +387,33 @@
 
     INSTALLED_APPS.append("debug_toolbar")
     MIDDLEWARE.append("debug_toolbar.middleware.DebugToolbarMiddleware")
+
+# STORAGE
+# Set to True and it will make use of the default django storage settings
```
cobusc: This comment is incorrect.

project/settings.py
```diff
@@ -374,3 +382,49 @@
 
     INSTALLED_APPS.append("debug_toolbar")
     MIDDLEWARE.append("debug_toolbar.middleware.DebugToolbarMiddleware")
+
+# STORAGE
+# Unless env.DEFAULT_STORAGE is set to false, this service will make use of the
```
cobusc: Should be "USE_DEFAULT_STORAGE" to match the line below

cobusc:
:+1: Code looks good. Just one comment that needs updating and Travis needs to pass.

PR #155, opened 2018-07-18
Handle UserSite creation in the 'user_accept_consent' callback

cobusc:
Dealing with it in the "site scope" handler was not sufficient, since a site that forget to add the site scope, or do not use it, will not be handled (as was the case with the Mangement Portal).


AltusBarry:
+1

PR #156, opened 2018-07-18
Feature/geinfra 67

AltusBarry:
There is some further enhancement ticketed: GEINFRA-119 

Not required for the functionality to work as expected.

### DIFF
--- a/authentication_service/api_helpers.py
+++ b/authentication_service/api_helpers.py
@@ -95,6 +95,7 @@ def get_invitation_data(invitation_id):
     # invitation_data.to_dict() causes a JSON serialisation error on the
     # datefields, there is no need for them at present.
     return {
+        "id": invitation_data.id,
         "first_name": invitation_data.first_name,
         "last_name": invitation_data.last_name,
         "email": invitation_data.email,
--- a/authentication_service/api_helpers.py
+++ b/authentication_service/api_helpers.py
@@ -92,8 +92,6 @@ def get_invitation_data(invitation_id):
     except AccessControlApiException as e:
         return {"error": True, "code": e.status}
 
-    # invitation_data.to_dict() causes a JSON serialisation error on the
-    # datefields, there is no need for them at present.
     return invitation_data.to_dict()
 
 
--- a/authentication_service/api_helpers.py
+++ b/authentication_service/api_helpers.py
@@ -94,13 +94,7 @@ def get_invitation_data(invitation_id):
 
     # invitation_data.to_dict() causes a JSON serialisation error on the
     # datefields, there is no need for them at present.
-    return {
-        "id": invitation_data.id,
-        "first_name": invitation_data.first_name,
-        "last_name": invitation_data.last_name,
-        "email": invitation_data.email,
-        "error": False
-    }
+    return invitation_data.to_dict()
 
 
 def invitation_redeem(invitation_id, user_id):
--- a/authentication_service/api_helpers.py
+++ b/authentication_service/api_helpers.py
@@ -92,3 +92,15 @@ def get_invitation_data(invitation_id):
     except AccessControlApiException as e:
         return {"error": True, "code": e.status}
     return invitation_data
+
+
+def invitation_redeem(invitation_id, user_id):
+    # API clients require uuid as a string.
+    user_id = str(user_id)
+    try:
+        redeem_data = settings.ACCESS_CONTROL_API.invitation_redeem(
+            invitation_id=invitation_id, user_id=user_id
+        )
+    except AccessControlApiException as e:
+        return {"error": True, "code": e.status}
+    return redeem_data
--- a/authentication_service/api_helpers.py
+++ b/authentication_service/api_helpers.py
@@ -1,3 +1,4 @@
+import types
 import logging
 
 from django.conf import settings
@@ -84,18 +85,24 @@ def get_user_site_role_labels_aggregated(user_id, client_id):
 
 
 def get_invitation_data(invitation_id):
-    # API clients require uuid as a string.
     try:
         invitation_data = settings.ACCESS_CONTROL_API.invitation_read(
             invitation_id
         )
     except AccessControlApiException as e:
         return {"error": True, "code": e.status}
-    return invitation_data
+
+    # invitation_data.to_dict() causes a JSON serialisation error on the
+    # datefields, there is no need for them at present.
+    return {
+        "first_name": invitation_data.first_name,
+        "last_name": invitation_data.last_name,
+        "email": invitation_data.email,
+        "error": False
+    }
 
 
 def invitation_redeem(invitation_id, user_id):
-    # API clients require uuid as a string.
     user_id = str(user_id)
     try:
         redeem_data = settings.ACCESS_CONTROL_API.invitation_redeem(
@@ -103,4 +110,4 @@ def invitation_redeem(invitation_id, user_id):
         )
     except AccessControlApiException as e:
         return {"error": True, "code": e.status}
-    return redeem_data
+    return {"error": False}
--- a/authentication_service/api_helpers.py
+++ b/authentication_service/api_helpers.py
@@ -81,3 +81,14 @@ def get_user_site_role_labels_aggregated(user_id, client_id):
     raise exceptions.ImproperlyConfigured(
         f"Site for client.id ({client_id}) not found."
     )
+
+
+def get_invitation_data(invitation_id):
+    # API clients require uuid as a string.
+    try:
+        invitation_data = settings.ACCESS_CONTROL_API.invitation_read(
+            invitation_id
+        )
+    except AccessControlApiException as e:
+        return {"error": True, "code": e.status}
+    return invitation_data
--- a/authentication_service/tests/test_views.py
+++ b/authentication_service/tests/test_views.py
@@ -601,7 +601,7 @@ class TestRegistrationView(TestCase):
         )
 
     def test_invite_tampered_signature(self):
-        invite_id = 1
+        invite_id = "8d81e01c-8a75-11e8-845e-0242ac120009"
         params = {
             "security": "high",
             "invitation": invite_id
@@ -639,7 +639,7 @@ class TestRegistrationView(TestCase):
     @patch("authentication_service.api_helpers.get_invitation_data")
     def test_invite_missing(self, mocked_get_invitation_data):
         mocked_get_invitation_data.return_value = {"error": True}
-        invite_id = 1
+        invite_id = "8d81e01c-8a75-11e8-845e-0242ac120009"
         params = {
             "security": "high",
             "invitation": invite_id
@@ -660,7 +660,7 @@ class TestRegistrationView(TestCase):
             "last_name": "same_as_above_but_surname",
             "email": "totallynotinvitation@email.com"
         }
-        invite_id = 1
+        invite_id = "8d81e01c-8a75-11e8-845e-0242ac120009"
         params = {
             "security": "high",
             "invitation": invite_id
@@ -697,7 +697,7 @@ class TestRegistrationView(TestCase):
         mocked_redeem.return_value = {
             "error": True
         }
-        invite_id = 1
+        invite_id = "8d81e01c-8a75-11e8-845e-0242ac120009"
         params = {
             "security": "high",
             "invitation": invite_id
--- a/authentication_service/tests/test_views.py
+++ b/authentication_service/tests/test_views.py
@@ -735,6 +735,46 @@ class TestRegistrationView(TestCase):
                 )
             self.assertContains(response, "Oops. You have")
 
+    @override_settings(ACCESS_CONTROL_API=MagicMock())
+    @patch("authentication_service.api_helpers.invitation_redeem")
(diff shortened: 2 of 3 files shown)


authentication_service/api_helpers.py
```diff
@@ -81,3 +82,33 @@ def get_user_site_role_labels_aggregated(user_id, client_id):
...
+
+def get_invitation_data(invitation_id):
+    try:
+        invitation_data = settings.ACCESS_CONTROL_API.invitation_read(
+            invitation_id
+        )
+    except AccessControlApiException as e:
+        return {"error": True, "code": e.status}
+
+    # invitation_data.to_dict() causes a JSON serialisation error on the
```
cobusc: It is important that we check that the invitation has not expired, so we cannot avoid passing it on.
`invitation_data.to_dict()` should work. It is not serialising to JSON, but rather de-serialising from JSON. If it cannot, then something is wrong with the data and we need to investigate.

authentication_service/tests/test_views.py
```diff
@@ -599,6 +602,138 @@ def setUpTestData(cls):
...
+            response = self.client.get(
+                reverse("registration"
+                ) + f"?invitation={invite_id}&signature={tampered_signature}",
+                follow=True
+            )
+
+        params = {
+            "invitation": invite_id
+        }
+        incorrect_security = signing.dumps(params, salt="invitation")
```
cobusc: This is actually "missing_security". I'd expect "incorrect_security" to contain `"security": "hithere"` or something like that.

authentication_service/tests/test_views.py
```diff
@@ -599,6 +602,138 @@ def setUpTestData(cls):
...
+        with self.assertTemplateUsed("authentication_service/message.html"):
+            response = self.client.get(
+                reverse("registration"
+                ) + f"?invitation={invite_id}&signature={incorrect_security}",
+                follow=True
+            )
+        params = {
+            "security": "high",
+        }
+        incorrect_signature = signing.dumps(params, salt="invitation")
```
cobusc: This is actually "missing_invitation", as per previous comment.

authentication_service/tests/test_views.py
```diff
@@ -599,6 +602,138 @@ def setUpTestData(cls):
...
+            updated_at=timezone.now()
+        )
+
+    def test_invite_tampered_signature(self):
+        invite_id = "8d81e01c-8a75-11e8-845e-0242ac120009"
+        params = {
+            "security": "high",
+            "invitation": invite_id
+        }
+        tampered_signature = signing.dumps(params, salt="invitation") + "m"
```
cobusc: Maybe also add a test for tampered content, which is what an attacker would typically do.

authentication_service/tests/test_views.py
```diff
@@ -599,6 +602,138 @@ def setUpTestData(cls):
...
+        with self.assertTemplateUsed("authentication_service/message.html"):
+            response = self.client.get(
+                reverse("registration"
+                ) + f"?invitation={invite_id}&signature={incorrect_signature}",
+                follow=True
+            )
+
+    @override_settings(ACCESS_CONTROL_API=MagicMock())
+    @patch("authentication_service.api_helpers.get_invitation_data")
+    def test_invite_missing(self, mocked_get_invitation_data):
```
cobusc: We need a test for an expired invitation too.

authentication_service/views.py
```diff
@@ -184,9 +181,57 @@ class RegistrationWizard(LanguageMixin, NamedUrlSessionWizardView):
...
+            # Check if signature is valid
+            try:
+                invitation_data = signing.loads(
+                    signature,
+                    salt="invitation",
+                )
+            except signing.BadSignature:
+                return error_response
+            invitation_id_matches = invitation_data.get("invitation") == invitation_id
+            if invitation_data.get("security") != "high" or \
```
cobusc: We don't need to check the value of `security`. The fact that it has not been tampered with should be good enough.

authentication_service/api_helpers.py
cobusc: This comment in the code is not relevant anymore and should be removed.

authentication_service/views.py
```diff
@@ -181,12 +185,112 @@ class RegistrationWizard(LanguageMixin, NamedUrlSessionWizardView):
...
+    instance_dict = {
+        "securityquestions": models.UserSecurityQuestion.objects.none()
+    }
     security = None
 
+    @cached_property
+    def inviter(self):
+        admin_id = self.storage.extra_data.get("invitation_data", {}).get("invitor_id")
+        try:
+            return get_user_model().objects.get(
```
cobusc: Rather do `USER_MODEL = get_user_model()` at the top of the module than calling the function repeatedly.

authentication_service/views.py
```diff
@@ -181,12 +185,112 @@ class RegistrationWizard(LanguageMixin, NamedUrlSessionWizardView):
...
+            )
+
+    def dispatch(self, request, *args, **kwargs):
+        dispatch = super(RegistrationWizard, self).dispatch(request, *args, **kwargs)
+
+        # Validate invitation and get data.
+        invitation = self.request.GET.get("invitation")
+        api_invitation = None
+        if invitation:
+            error_response = render(
```
cobusc: Please refactor this into a helper function that can be called when required. This function is a bit long and the construction of this response is unnecessary when everything is OK.

authentication_service/views.py
```diff
@@ -181,12 +185,112 @@ class RegistrationWizard(LanguageMixin, NamedUrlSessionWizardView):
...
+            # Storage value needed for the inviter property
+            self.storage.extra_data["invitation_data"] = api_invitation
+            self.storage.extra_data["invitation_setup"] = invitation_data
+
+            # Set check if organisation exists and set cached property in the
+            # process.
+            self.organisation
+            if expires_at < timezone.now():
+                inviter = self.inviter
+                return render(
```
cobusc: Please refactor this into a helper function that can be called when required. This function is a bit long.
Something like `self._render_expired_invitation_page(request, invitor)`

authentication_service/views.py
```diff
@@ -197,18 +301,26 @@ def get_form_initial(self, step):
...
+            "required": self.request.GET.getlist("requires"),
+            "hidden": self.request.GET.getlist("hide"),
+            "question_ids": self.request.GET.getlist("question_ids", [])
+        }
+
+        custom_kwargs.update(
+            self.storage.extra_data.get("invitation_setup", {})
+        )
+
+        # TODO: Can be refactored later
```
cobusc: Now is a good time :)

authentication_service/views.py
```diff
@@ -181,12 +185,112 @@ class RegistrationWizard(LanguageMixin, NamedUrlSessionWizardView):
...
+            # session storage.
+            api_invitation.pop("created_at")
+            api_invitation.pop("updated_at")
+            expires_at = api_invitation.pop("expires_at")
+
+            # Storage value needed for the inviter property
+            self.storage.extra_data["invitation_data"] = api_invitation
+            self.storage.extra_data["invitation_setup"] = invitation_data
+
+            # Set check if organisation exists and set cached property in the
```
cobusc: Comment should probably starts like this: "Check..."

authentication_service/views.py
```diff
@@ -181,12 +185,112 @@ class RegistrationWizard(LanguageMixin, NamedUrlSessionWizardView):
...
+            # Do some validation with the invitation data
+            if api_invitation.get("error") is True:
+                return error_response
+            # Prevents needing to manipulate data before being saved to
+            # session storage.
+            api_invitation.pop("created_at")
+            api_invitation.pop("updated_at")
+            expires_at = api_invitation.pop("expires_at")
+
+            # Storage value needed for the inviter property
```
cobusc: ..and organisation

authentication_service/views.py
```diff
@@ -276,6 +383,35 @@ def done(self, form_list, **kwargs):
                 data["language_code"] = self.language
                 question = models.UserSecurityQuestion.objects.create(**data)
 
+        invitation = self.storage.extra_data.get("invitation_data")
+        if invitation:
+            inviter = self.inviter
+            error_response = render(
```
cobusc: Refactor to helper function, please.

authentication_service/views.py
```diff
@@ -181,12 +185,112 @@ class RegistrationWizard(LanguageMixin, NamedUrlSessionWizardView):
...
+            api_invitation.pop("updated_at")
+            expires_at = api_invitation.pop("expires_at")
+
+            # Storage value needed for the inviter property
+            self.storage.extra_data["invitation_data"] = api_invitation
+            self.storage.extra_data["invitation_setup"] = invitation_data
+
+            # Set check if organisation exists and set cached property in the
+            # process.
+            self.organisation
```
cobusc: I _really_ dislike this, so much. I'd much prefer ditching the @cached_property function and maybe doing:
```
try:
    self.invitation = models.Organisation.objects.get(id=api_invitation["invitation_id"])
except models.Organisation.DoesNotExist:
    return render_unknown_organisation_page(request, ...)
```
Same for inviter cached property. Is it really necessary?

AltusBarry: Putting things on the instance with session wizards is liable to break or leak into other instances. cached_property is safer.

cobusc: A cached property does exactly that (puts something on the instance)! 
```
The @cached_property decorator caches the result of a method with a single self argument as a property. The cached result will persist as long as the instance does, so if the instance is passed around and the function subsequently invoked, the cached result will be returned.
```
https://docs.djangoproject.com/en/1.11/ref/utils/#django.utils.functional.cached_property
My original comment still stands.

authentication_service/views.py
AltusBarry: This specific response should only ever occur once, moved it to just return the render. Instead of instantiating it unnecessarily.

authentication_service/api_helpers.py
cobusc: You still need to remove this comment.

authentication_service/tests/test_views.py
```diff
@@ -599,6 +603,242 @@ def setUpTestData(cls):
...
+                        "userdata-birth_date": "2000-01-01",
+                        "userdata-terms": True,
+                        "userdata-email": "email@email.com",
+                    },
+                    follow=True
+                )
+        self.assertContains(response, "Congratulations, you have successfully")
+
+        user = get_user_model().objects.get(username="AmazingInviteUser")
+        self.assertEqual(user.organisation, self.organisation)
```
cobusc: Bonus points: Check that `settings.ACCESS_CONTROL_API.invitation_read` was called.

authentication_service/views.py
```diff
@@ -169,24 +172,131 @@ def post(self, *args, **kwargs):
...
+            api_invitation = api_helpers.get_invitation_data(
+                invitation_data.pop("invitation_id")
+            )
+
+            # Do some validation with the invitation data
+            if api_invitation.get("error") is True:
+                return self._url_signature_error()
+            # Prevents needing to manipulate data before being saved to
+            # session storage.
+            api_invitation.pop("created_at")
```
cobusc: Just FYI `del api_invitation["created_at"]` would also have worked.

authentication_service/views.py
```diff
@@ -196,19 +306,21 @@ def get_form_initial(self, step):
...
-            self.storage.extra_data["required"] = required
-        hidden = self.request.GET.getlist("hide")
-        if hidden:
-            self.storage.extra_data["hidden"] = hidden
-        question_ids = self.request.GET.getlist("question_ids", [])
-        if question_ids:
-            self.storage.extra_data["question_ids"] = question_ids
+        for key, value in custom_kwargs.items():
+            if value:
+                self.storage.extra_data[key] = custom_kwargs[key]
```
cobusc: `self.storage.extra_data[key] = value` No need to look up the value again

authentication_service/views.py
```diff
@@ -254,12 +357,7 @@ def done(self, form_list, **kwargs):
         pwd = self.get_all_cleaned_data()["password1"]
 
         # Save user model
-        user =  kwargs["form_dict"]["userdata"].save()
-
-        # GE-1065 requires ALL users to be logged in
-        new_user = authenticate(username=username,
-                                password=pwd)
-        login(self.request, new_user)
+        user = kwargs["form_dict"]["userdata"].save()
```
cobusc: Can `userdata` be updated to contain the organisation?

authentication_service/tests/test_views.py
cobusc: Did you add a test for this?

authentication_service/views.py
AltusBarry: facepalm moment, thanks

authentication_service/tests/test_views.py
AltusBarry: ```671     def test_expire(self):```

cobusc: Thanks

authentication_service/views.py
cobusc: :+1:

cobusc:
:+1:

cobusc:
:+1:

PR #158, opened 2018-07-23
Only API key authentication should be allowed for API calls

cobusc:
# Background
API calls were allowed for users that are logged in or used basic HTTP auth. This should not be allowed.

# What was done
Restrict API authentication to only look at HTTP header auth. The `utils.py` file that gets generated by the `swagger-django-generator` is sent to `/dev/null` so that we preserve the `utils.py` that only allows the HTTP header auth.
Tests were updated to use the auth header.


AltusBarry:
+1

PR #159, opened 2018-07-23
feature/GEINFRA-137: Move login logic to later in wizard, guards agai…

AltusBarry:
…nst wizard session KeyErrors

### DIFF
--- a/authentication_service/views.py
+++ b/authentication_service/views.py
@@ -412,27 +412,11 @@ class RegistrationWizard(LanguageMixin, NamedUrlSessionWizardView):
         return self.get_success_response()
 
     def render_done(self, form, **kwargs):
-        # This validation is done in the base wizard render_done as well.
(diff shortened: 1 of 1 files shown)


authentication_service/views.py
```diff
@@ -406,12 +407,36 @@ def done(self, form_list, **kwargs):
...
+            form_obj = self.get_form(
+                step=form_key,
+                data=self.storage.get_step_data(form_key),
+                files=self.storage.get_step_files(form_key)
+            )
+            if not form_obj.is_valid():
+                return self.render_revalidation_failure(form_key, form_obj, **kwargs)
+            final_forms[form_key] = form_obj
+
+        # Call super that validates forms again and resets storage.
```
cobusc: If we are overriding the base method (instead of extending it), can't we skip the call to `super()` and rather do the reset storage ourselves here?

PR #161, opened 2018-07-31
Added API definition for the 'request_user_deletion' end-point

cobusc:
This is just the Swagger definition and related stubs. The function has not been implemented.
Note: I moved the implementation of the `purge_expired_invitations` function to match the position in the generated stubs file.


AltusBarry:
+1

PR #162, opened 2018-08-02
Delete user data via asynchronous (celery) task

cobusc:


authentication_service/tasks.py
```diff
@@ -178,7 +182,99 @@ def purge_expired_invitations(cutoff_date):
...
+    # Delete Access Control data
+    result = operational_api.delete_user_data(user_id)
+    logger.debug(f"{result['amount']} rows deleted from Access Control")
+
+    # Delete User Data Store data and set the "deleted_at" value
+    # of the DeletedUser entity.
+    result = user_data_store_api.delete_user_data(user_id)
+    logger.debug(f"{result['amount']} rows deleted from User Data Store")
+
+    user_data_store_api.deleteduser_update(
```
AltusBarry: Should this call not perhaps be after `user.delete()` in the event delete() causes some form of error?

authentication_service/tests/test_tasks.py
```diff
@@ -183,3 +183,79 @@ def test_purge_expired_invitation_task(self):
...
+
+class DeleteUserAndData(TestCase):
+
+    @classmethod
+    def setUpTestData(cls):
+        user_model = get_user_model()
+
+        cls.organisation = Organisation.objects.create(
+            id=1,
+            name=f"test unit",
```
AltusBarry: Doesn't need to be an f string

authentication_service/tests/test_tasks.py
```diff
@@ -183,3 +183,79 @@ def test_purge_expired_invitation_task(self):
...
+            description="Description"
+        )
+
+        cls.deleter = user_model.objects.create(
+            username="deleter",
+            first_name="Del",
+            last_name="Eter",
+            email="deleter@example.com",
+            is_superuser=False,
```
AltusBarry: The `is_superuser` and `is_staff` values are False by default.

authentication_service/tasks.py
```diff
@@ -178,7 +182,99 @@ def purge_expired_invitations(cutoff_date):
...
+    return settings.AC_OPERATIONAL_API.purge_expired_invitations(
         cutoff_date=cutoff_date
     )
 
+
+@task(name="delete_user_and_data_task",
+      default_retry_delay=5 * 60,
+      autoretry_for=(AccessControlApiException, UserDataStoreApiException),
+      retry_backoff=True,
+      retry_backoff_max=600,
```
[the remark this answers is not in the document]
cobusc: Yeah, I add it so that this option is visible to anyone who may not know that it exists and want to tweak it.

authentication_service/tasks.py
```diff
@@ -178,7 +182,99 @@ def purge_expired_invitations(cutoff_date):
...
+      retry_jitter=True)
```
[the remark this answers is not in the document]
cobusc: See above

authentication_service/tasks.py
cobusc: As discussed, if the order is changed like you suggest, the function will not be idempotent anymore.

authentication_service/tests/test_tasks.py
cobusc: You've pointed out the default values so often that I want to counter with: "Explicit is better than implicit." However, in this particular case I will remove it since it the values of those fields are not relevant.

authentication_service/tasks.py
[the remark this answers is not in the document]
AltusBarry: :+1:

authentication_service/tasks.py
AltusBarry: :+1:

AltusBarry:
+1

PR #175, opened 2018-08-27
Feature/geinfra 119

AltusBarry:
Make use of underlying ModelForm used by registration to save the organisation.

- Removed saving logic from the view
- Display the organisation the user has been invited to by manipulating the default ModelChoice field attrs.

### DIFF
--- a/authentication_service/forms.py
+++ b/authentication_service/forms.py
@@ -102,13 +102,13 @@ class RegistrationForm(UserCreationForm):
             field.label = _("Oragnisation user has been invited to:")
 
             # Override widget templates
-            widget.template_name = "authentication_service/widgets/select.html"
+            widget.template_name = "authentication_service/widgets/options_display_only.html"
 
-            # Dango form renderers patches the BoundField.css_classes() method
+            # Dango form renderer patches the BoundField.css_classes() method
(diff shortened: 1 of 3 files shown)


authentication_service/forms.py
```diff
@@ -66,20 +66,55 @@ class RegistrationForm(UserCreationForm):
...
+            # Override widget templates
+            widget.template_name = "authentication_service/widgets/options_display_only.html"
+
+            # Dango form renderer patches the BoundField.css_classes() method
+            # to include the bound field name as part of the classes. This is a
+            # special case where the underlying functionality of the
+            # ModelChoice field is to be used, without any of its front end
+            # components or styling.
+            self["organisation"].field.__class__.__name__ = \
+                "OverridenBoundFieldClassName"
```
cobusc: Is this an arbitrary value you have chosen?

AltusBarry: Yes

cobusc: Please mention this explicitly in a comment. It looks like a legitimate name, so the expectation is that it carries some meaning. If it was something like "ARB_NAME_TO_AVOID_PATCHING", then it would have been more explicit.

cobusc:
:+1:

PR #176, opened 2018-08-27
feature/BoundFieldStyleFix: Bit the bullet and added a custom field a…

AltusBarry:
…nd widget.

image


authentication_service/fields.py
```diff
@@ -0,0 +1,42 @@
+from django.forms.fields import Field
+from django.utils.text import mark_safe
+from django.utils.translation import ugettext as _
+
+from authentication_service import widgets
+
+
+class ParagraphField(Field):
```
cobusc: Nice

cobusc:
:+1:

PR #178, opened 2018-09-04
feature/GEINFRA-245: Added fix and test for sentry error

AltusBarry:


authentication_service/forms.py
```diff
@@ -674,8 +674,8 @@ def __init__(self, *args, **kwargs):
 
     def clean(self):
         for question in self.questions:
-            if not self.cleaned_data["question_%s" % question.id]:
-                raise ValidationError(_("Please enter your answer."))
+            if not self.cleaned_data.get("question_%s" % question.id, None):
+                raise ValidationError(_("Please enter all the answers."))
```
DavidOberholzer: did you write this copy? :P

AltusBarry: Indeed. I can also see why you would ask that. Let me update it. :sweat_smile:

PR #179, opened 2018-09-10
Feature/geinfra 29

AltusBarry:


project/settings.py
```diff
@@ -66,6 +66,7 @@
     ("my", _("Burmese")),
     ("ny", _("Chichewa")),
     ("prs", _("Dari")),
+    ("en-you", _("Springster english youth")),
```
cobusc: Can we capitalise these? At least "English", but maybe also "Youth".

cobusc:
:+1:

PR #183, opened 2018-09-25
Feature/geinfra 274

AltusBarry:


authentication_service/management/commands/demo_content.py
```diff
@@ -177,6 +177,25 @@ def handle(self, *args, **options):
             "Created" if created else "Updated", c.client_id
         )))
 
+        c, created = Client.objects.update_or_create(
+            client_id="core_data_ingestion",
+            defaults={
+                "name": "Data ingestion service",
+                "client_secret": "core_data_ingestion_secret",
```
DavidOberholzer: Title case like the rest maybe: `Data Ingestion Service`

PR #184, opened 2018-09-28
Feature/geinfra 52

AltusBarry:
Add ability to request client creation


authentication_service/request_client/models.py
```diff
@@ -0,0 +1,123 @@
...
+from authentication_service import tasks
+
+from project.settings import MediaStorage
+from oidc_provider.models import (
+    CLIENT_TYPE_CHOICES,
+    RESPONSE_TYPE_CHOICES,
+    JWT_ALGS
+)
+
+ENVIRONMENT_CHOICES = (
```
DavidOberholzer: This can go in `constants.py` I think

AltusBarry: At this stage the values are not used anywhere else, so in the module that uses it is fine.

PR #186, opened 2018-10-02
feature/GEINFRA-293: Streamline decorator a bit

AltusBarry:


authentication_service/user_migration/forms.py
```diff
@@ -57,6 +60,7 @@ def clean_password2(self):
                 _("Password needs to be at least 4 characters long.")
             )
         return password2
+UserDataForm.__init__ = required_form_fields_label_alter(UserDataForm.__init__)
```
cobusc: Did you try defining the class as 
```python
@method_decorator(required_form_fields_label_alter, name='__init__')
class UserDataForm(forms.Form):
    ....
```
?

AltusBarry: Yeah I did. the decorator args and kwargs are completely empty. Then attempting to address that it broke the decorator for the other use cases. As this ticket was timeboxed, I had to revert and make use of this method.

cobusc: OK

PR #187, opened 2018-10-11
Feature/geinfra 42

AltusBarry:
Add management command to email a csv with some user info, based on defined filters

### DIFF
--- a/authentication_service/management/commands/user_login_report.py
+++ b/authentication_service/management/commands/user_login_report.py
@@ -97,7 +97,7 @@ class Command(BaseCommand):
 
     def handle(self, *args, **options):
         self.stdout.write(self.style.SUCCESS("Fetching users..."))
-        recipients = settings.USER_LOGIN_REQUEST_EMAIL
+        recipients = settings.USER_LOGIN_REQUEST_EMAILS
         now = timezone.now()
 
         # Append the email provided via the client arg
--- a/authentication_service/management/commands/user_login_report.py
+++ b/authentication_service/management/commands/user_login_report.py
@@ -3,14 +3,18 @@ import csv
 import datetime
 import io
 
+from django.conf import settings
 from django.contrib.auth import get_user_model
 from django.core.exceptions import ValidationError
+from django.core.mail import EmailMultiAlternatives
 from django.core.management.base import BaseCommand
+from django.core.management.base import CommandError
 from django.core.validators import validate_email
 from django.utils import timezone
-from django.core.mail import EmailMultiAlternatives
 
 from dateutil.relativedelta import relativedelta
+
+from authentication_service.tasks import FROM_EMAIL
 
 
 def date_range(string):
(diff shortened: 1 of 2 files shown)


authentication_service/management/commands/user_login_report.py
```diff
@@ -0,0 +1,165 @@
...
+                " Everything after a date till today: yyyy/mmm/dd,*"
+            )
+        )
+
+    def handle(self, *args, **options):
+        self.stdout.write(self.style.SUCCESS("Fetching users..."))
+        recipients = settings.USER_LOGIN_REQUEST_EMAIL
+        now = timezone.now()
+
+        # Append the email provided via the clie arg
```
cobusc: Typo: "clie"

authentication_service/management/commands/user_login_report.py
```diff
@@ -0,0 +1,165 @@
...
+        if options.get("additional_email") is not None:
```
cobusc: Can we have multiple additional emails? Then we can do something like:
```
additional_emails = options.get("additional_emails", [])
recipients.extend(additional_emails)
```

authentication_service/management/commands/user_login_report.py
```diff
@@ -0,0 +1,165 @@
...
+            flag = True if options["user_status"] == "active" else False
+            filter_kwargs["is_active"] = flag
+
+        # Filter the users
+        users = user_model.objects.filter(
+            **{k: v for k, v in filter_kwargs.items()}
+        )
+
+        # Write a csv to memory
+        file = io.StringIO()
```
cobusc: We may have to consider writing to disk in the future. If we have 100ks users then this will be a problem.

authentication_service/management/commands/user_login_report.py
```diff
@@ -0,0 +1,165 @@
...
+
+        # Add filter for active flag
+        if options.get("user_status") != "both":
+            flag = True if options["user_status"] == "active" else False
+            filter_kwargs["is_active"] = flag
+
+        # Filter the users
+        users = user_model.objects.filter(
+            **{k: v for k, v in filter_kwargs.items()}
+        )
```
cobusc: Only pull out the fields that we are going to use in the CSV using `values_list`. This will help us mitigate large resultset issues.

project/settings.py
```diff
@@ -21,6 +21,9 @@
 
 # NOTE: If the value is not set, the celery task will error.
 CLIENT_REQUEST_EMAIL = env.list("CLIENT_REQUEST_EMAIL", [])
+
+# authentication_service/management/commands/user_login_report.py
+USER_LOGIN_REQUEST_EMAIL = env.list("USER_LOGIN_REQUEST_EMAIL", [])
```
cobusc: I see it is a list. Use plural name, please.

cobusc:
:+1:

PR #188, opened 2018-10-12
feature/GEINFRA-305: Remove unique 404 path metrics logging from metr…

AltusBarry:
…ic middleware

### DIFF
--- a/authentication_service/middleware.py
+++ b/authentication_service/middleware.py
@@ -365,7 +365,12 @@ class MetricMiddleware(object):
     def __call__(self, request):
         start_time = time.time()
         response = self.get_response(request)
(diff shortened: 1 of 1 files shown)


authentication_service/middleware.py
```diff
@@ -365,7 +365,12 @@ def __init__(self, get_response):
     def __call__(self, request):
         start_time = time.time()
         response = self.get_response(request)
-        H.labels(path_prefix="/".join(request.path.split("/")[:4]),
+        if response.status_code == 404:
```
cobusc: How about:
```
path_prefix = "not_found" if response.status_code == 404 else "/".join(request.path.split("/")[:4])
H.labels(path_prefix=path_prefix, method=request.method, status=response.status_code).observe(time.time()-start_time)
```

PR #190, opened 2018-11-14
feature/GEINFRA-351: Show success page with option to go back to clie…

AltusBarry:
…nt on pwd reset success pages

Also fix a case where a session variable was not getting correctly cleared.

screenshot_2018-11-14 password update success


cobusc:
:+1: Just check why the PR build failed, please.

PR #191, opened 2018-11-26
Update Springster and NiNyampinga themes

KaitCrawford:


markupguy:
@KaitCrawford I seemed to have approved it already. It's just CSS, so fire away :)

PR #193, opened 2018-11-29
Kinesis startup change

cobusc:
# Background
We had a problem on QA where the application would not start up. The issue was reproduced in the docker compose environment.

# What was done
It was noticed that the migration process that runs before the application is started, never finishes.
This behaviour was seen before when the Kinesis producer process has been created. The fix was to only enable the Kinesis producer when
[1] the USE_KINESIS_PRODUCER environment variable is true AND
[2] the BUILDER environment variable is NOT true.


AltusBarry:
Rather update https://github.com/girleffect/core-authentication-service/blob/develop/scripts/startDjango.sh
with
```
-BUILDER=true python manage.py migrate --noinput
+USE_KINESIS_PRODUCER=false python manage.py migrate --noinput

-python manage.py runserver 0.0.0.0:8000
+tini -- /scripts/django-entrypoint.sh project.wsgi:application
```

AltusBarry:
:+1:

PR #200, opened 2019-04-29
Themes setup assets update

Mitso:


Mitso:
@codiebeulaine 
Please review
Themes folder updated, added Chaajaa, Yegna

PR #211, opened 2019-06-20
update static root settings

moh-moola:


project/settings.py
```diff
@@ -404,7 +404,7 @@
     )
 
 # SECURITY WARNING: don't run with debug turned on in production!
-DEBUG = env.bool("DEBUG", False)
+DEBUG = env.bool("DEBUG", True)
```
codiebeulaine: ?

moh-moola: By default the Debug mode should be set to True and explicitly set to False in production, but this is not the case for the Auth Service in prod

moh-moola:
@codiebeulaine updated the above issue

codiebeulaine:
👍

